PR #14, opened 2019-09-19
Disabled STM32F745 based targets until this Unified Target is fixed.

mikeller:
Mitigation for betaflight/betaflight#8908.


DieHertz:
Are you sure about clracing? Mine was f722, but maybe it's an older version or something

IvoFPV:
@mikeller CLRACING F7 is F722 we fixed that recently #12, shouldn't be disabled.

mikeller:
Ha, turns out I'd done the fix on a stale checkout of the repository - mentally these are still _inside_ the betaflight repository and don't need a separate `git pull`> :confused: 

But anyhow, as we've got a fix now (betaflight/betaflight#8914), I don't think pulling them out and then putting them in again is needed.

mikeller:
Closing this as betaflight/betaflight#8914 was merged.

PR #33, opened 2019-09-30
update unified target AIRBOTF7

x4FF3:
hi,

update to the unified target to add the missing dma.

KR


mikeller:
@x4FF3 : Are you going to be managing the target configurations for Airbot? If so I can give you the rights required to approve pull requests for these targets.

PR #38, opened 2019-10-01
Removed extensions for cleaned up files from '.gitattributes'.

mikeller:


etracer65:
Looks like this commit updated an extra unintended file.

mikeller:
@etracer65: Yes, this is the normalisation of one file that had snuck in with wrong line endings and got me to discover the problem.

PR #40, opened 2019-10-02
add custom settings for MATEK targets

MATEKSYS:


configs/default/MTKS-MATEKF405CTR.config
```diff
@@ -105,11 +105,21 @@ dma pin A01 0
...
+
+# serial
+serial 1 64 115200 57600 0 115200
 
 # master
+set serialrx_provider = SBUS
+set motor_pwm_protocol = DSHOT600
 set mag_bustype = I2C
 set mag_i2c_device = 1
+set mag_hardware = NONE
```
McGiverGim: Is there a reason to not let this kind of configurations as AUTO? It will be easy if the user connects an external MAG.

MATEKSYS: Just make a custom setting for most users.

mikeller: @MATEKSYS: While it is up to you how you set up your targets, I am with @McGiverGim in wondering what benefits you see in not enabling auto-detection for the magnetometer?
Since most of your targets have an on board barometer that shares the I2C bus with the I2C pads for an external magnetometer, it is not a case where users could easily re-use these pins for something else anyway. Leaving `mag_hardware` on the default `auto` means there is one less thing to configure for users connecting a magnetometer, and it does not make a difference for users _not_ connecting one.

MATEKSYS: In my memory most users don't use magnetometer with GPS-rescue-mode and they would prefer turn it off,  Just want to save one click for them.

McGiverGim: I prefer to use MAG with GPS Rescue, I have quads with and without it. But to me it seems more precise. 

But if you want to use only GPS when in GPS Rescue mode, then there is another setting `gps_rescue_use_mag`.

mikeller: @MATEKSYS:

> In my memory most users don't use magnetometer with GPS-rescue-mode and they would prefer turn it off, Just want to save one click for them.

My point is that there is no need for users to turn off the magnetometer when it is set to `auto` - `auto` and `none` act the same way when no magnetometer is present.

Docteh: I have the opposite problem, friends turning on the OSD compass without a MAG 😠

mikeller: @Docteh: I think it 'works' based on IMU data in this case. Do you think it would be better if we signaled 'not supported' if no magnetometer is detected?

McGiverGim: After GPS Rescue maybe it works too if a GPS and not a MAG is detected? I'm not sure if this was changed. The GPS Home arrow was.

mikeller: @McGiverGim: GPS is likely to be even more dodgy when it comes to overlaying a compass direction onto the OSD, as it can only use direction inferred from movement, and even then has to make the potentially wrong assumption that the craft's camera angle is aligned with the craft's direction of movement, which it isn't.

PR #41, opened 2019-10-02
Add LuxMiniF7

x4FF3:
Hi,

unified target for the LuxMiniF7.

KR


mikeller:
@x4FF3: Are you going to be the maintainer for Airbot's target configurations? I can set you up so you can approve pull requests for Airbot target configurations.

x4FF3:
@mikeller yes ill maintain them.

mikeller:
@x4FF3: Is this hardware actually going to be sold and supported by Airbot? If not, the manufacturer id should not be 'AIRB', which in a future version of configurator will be used to direct users to Airbot's website for support.

x4FF3:
@mikeller ill clarify it.

PR #45, opened 2019-10-03
Unified target for AIKONF7

DieHertz:
Should old-fashioned target and docs be added separately?

### DIFF
--- a/configs/default/AIKO-AIKONF7.config
+++ b/configs/default/AIKO-AIKONF7.config
@@ -110,4 +110,4 @@ set flash_spi_bus = 3
 set gyro_1_bustype = SPI
 set gyro_1_spibus = 1
 set gyro_1_sensor_align = CW0
-set pinio_box = 39,0,0,0
+set pinio_box = 40,0,0,0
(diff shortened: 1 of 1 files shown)


etracer65:
If this target is intended to have VTX power control using `pinio`, then don't use mode 39 (`VTX_PIT_MODE`) to control it. Use mode 40 (`USER1`) instead. Otherwise the only way to power the VTX is to have pit mode active.

DieHertz:
Thanks, I'll change that, the intention is to have full control over VTX power.

mikeller:
@DieHertz: Have a re-read of https://github.com/betaflight/betaflight#new-requirements-for-the-submission-of-new-and-updated-targets, it has been updated recently. Updates to non-unified targets (and board documentation) are not required any more, but for now we still accept them.
We'll move away from having board documentation hosted in our repository at some point, probably to a system where we keep metadata for each board including an URL for a board information page with each Unified Target configuration, and make it available from within configurator.

Different question: Are you going to be the maintainer for AIKON's targets? If so I can assign you as a reviewer for these target definitions.

DieHertz:
Yes, you can also add me for tmotor and spedix.
Is the mechanism to flash configs from BFC already implemented?

mikeller:
@DieHertz: Done. And yes, it is included in the 10.6.0 release candidates.

PR #51, opened 2019-10-11
Add custom config for homebrew CUST-SIRMIXALOT.config

jrowe88:


configs/default/CUST-SIRMIXALOT.config
```diff
@@ -0,0 +1,187 @@
+# 
+
+# diff all bare
+
+# version
+# Betaflight / STM32F7X2 (S7X2) 4.1.0 Sep 25 2019 / 14:09:57 (08e8afa09) MSP API: 1.42
```
mikeller: Have you tested this?

The banner starting in `# Betaflight` has to be the first line in the file, or this will not work.

configs/default/CUST-SIRMIXALOT.config
```diff
@@ -0,0 +1,187 @@
...
+aux 3 15 5 1300 2100 0 0
+aux 4 26 0 900 2100 0 0
+aux 5 33 2 1300 1700 0 0
+aux 6 34 3 1300 1700 0 0
+aux 7 36 6 1700 2100 0 0
+
+# vtxtable
+vtxtable bands 5
+vtxtable channels 8
+vtxtable band 1 BANDA    A FACTORY 5865 5845 5825 5805 5785 5765 5745 5725
```
mikeller: This seems to be making this target illegal to be used in pretty much every country.

There is no point in adding targets that cannot be used without breaking the law.

configs/default/CUST-SIRMIXALOT.config
```diff
@@ -0,0 +1,187 @@
...
+vtxtable band 2 BANDB    B FACTORY 5733 5752 5771 5790 5809 5828 5847 5866
+vtxtable band 3 BANDE    E FACTORY 5705 5685 5665 5645 5885 5905 5925 5945
+vtxtable band 4 BANDF    F FACTORY 5740 5760 5780 5800 5820 5840 5860 5880
+vtxtable band 5 RACE     R FACTORY 5658 5695 5732 5769 5806 5843 5880 5917
+vtxtable powerlevels 4
+vtxtable powervalues 25 200 500 800
+vtxtable powerlabels 25 200 500 800
+
+# master
+set acc_calibration = 24,-24,38
```
mikeller: This should not be in here, as it is specific to the acc chip on one board.

configs/default/CUST-SIRMIXALOT.config
```diff
@@ -0,0 +1,187 @@
...
+
+rateprofile 2
+
+rateprofile 3
+
+rateprofile 4
+
+rateprofile 5
+
+#
```
mikeller: These should not be in here, as they are empty.

configs/default/CUST-SIRMIXALOT.config
jrowe88: No; sorry, was literally following the instructions and didn't remove the first lines.

configs/default/CUST-SIRMIXALOT.config
jrowe88: Again, used diff command and it splatted that out...I'll remove it.

configs/default/CUST-SIRMIXALOT.config
jrowe88: Maybe I'll revise the instructions on how to make a config....

configs/default/CUST-SIRMIXALOT.config
jrowe88: That's not my intention.  I'm not sure I follow you though as from a few different sources those are the frequencies for Band A.

configs/default/CUST-SIRMIXALOT.config
mikeller: It looks like you forgot to reset your custom target to defaults in step 1 (https://github.com/betaflight/betaflight/blob/master/docs/TargetMaintenance/CreatingAUnifiedTarget.md#1-flash-the-firmware-onto-your-board):

- verify that your board is properly reset to `defaults`: The output of a `diff hardware` in CLI should show `board_name`, `manufacturer_id`, and lines starting with a `#`.

jrowe88:
@mikeller - updates pushed.  I would like to understand/be educated on your comments for VTX bands.  I removed them, but thought they were the valid frequencies for Band A.

x4FF3:
>  updates pushed. I would like to understand/be educated on your comments for VTX bands. I removed them, but thought they were the valid frequencies for Band A.

@jrowe88 
as not every frequency is legal in every country on this planet, targets should not include this setting as default.

Docteh:
What is this config intended to be used for? is there a "SIRMIXALOT" board?

jrowe88:
@Docteh, yes, there is a SirMixALot board.

> What is this config intended to be used for? is there a "SIRMIXALOT" board?

@Docteh - yes, there is a SirMixALot board.

PR #59, opened 2019-10-17
Add Airbot OMNINXT4 unified config file

McGiverGim:
Adds Airbot OMNINXT4 unified target. 

I think this target is not commercially available, but I had a sample and created it as stated here: https://github.com/betaflight/betaflight/issues/9044


etracer65:
There are some differences as compared to the legacy target. The UARTs should be defaulted (with `serial` commands) as follows (from config.c):
```
    { SERIAL_PORT_USART1, FUNCTION_RX_SERIAL },
    { SERIAL_PORT_UART5,  FUNCTION_ESC_SENSOR },
```
And from target.h we should have `feature RX_SERIAL` based on:
```
#define DEFAULT_RX_FEATURE      FEATURE_RX_SERIAL
```

McGiverGim:
True, I was looking for the RX_SERIAL feature in the Configurator without luck... better to go to CLI :)

Fixed as per review...

PR #62, opened 2019-10-17
Add HAKRCF405 unified config

IvoFPV:


etracer65:
Should also have:
```
feature OSD
```

etracer65:
Looks like the same kind of changes should apply to the `DALRCF405` and `FOXEERF405` targets as well.

IvoFPV:
@etracer65 Yeah, updated them all. 😉

PR #63, opened 2019-10-17
Add DALRCF722DUAL unified target

etracer65:
Created using a real DALRCF722DUAL board.

### DIFF
--- a/configs/default/DALR-DALRCF722DUAL.config
+++ b/configs/default/DALR-DALRCF722DUAL.config
@@ -1,0 +1,120 @@
+# Betaflight / STM32F7X2 (S7X2) 4.1.0 Oct 16 2019 / 11:58:45 (c37a7c91a) MSP API: 1.42
+
+board_name DALRCF722DUAL
+manufacturer_id DALR
+
+# resources
+resource BEEPER 1 C13
+resource MOTOR 1 C06
+resource MOTOR 2 C07
+resource MOTOR 3 C08
+resource MOTOR 4 C09
+resource MOTOR 5 A08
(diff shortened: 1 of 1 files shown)


configs/default/DALR-DALRCF722DUAL
```diff
@@ -0,0 +1,124 @@
+# Betaflight / STM32F7X2 (S7X2) 4.1.0 Oct 16 2019 / 11:58:45 (c37a7c91a) MSP API: 1.42
```
mikeller: Heh, that file name. :wink:

mikeller:
Also, checks are working. :+1:

configs/default/DALR-DALRCF722DUAL
etracer65: Oops! Fixed.

configs/default/DALR-DALRCF722DUAL.config
```diff
@@ -0,0 +1,123 @@
...
+set max7456_spi_bus = 2
+set dashboard_i2c_bus = 1
+set flash_spi_bus = 3
+set gyro_1_bustype = SPI
+set gyro_1_spibus = 1
+set gyro_1_sensor_align = CW180
+set gyro_1_align_yaw = 1800
+set gyro_2_spibus = 1
+set battery_meter = ADC
+set current_meter = ADC
```
IvoFPV: @etracer65 These are repeated twice.

etracer65: Ahhh, you're right - thanks. Fixed.

PR #65, opened 2019-10-17
Fix and Update AlienFlightNGF7 unified configs

MJ666:


MJ666:
Fixed the rx_spi_protocol setting. The related FC is still in prototyping phase and we are using an ready made CC2500 board. This module supports diversity, has an LNA/PA and can be soldered to the FC. https://www.aliexpress.com/i/32945151938.html

mikeller:
@MJ666: Good stuff!

I hope the long standing issues with FrSky X telemetry should finally be fixed with betaflight/betaflight#9010, so that's ready for prime time now.

Re PA/LNA, most of the currently used micro all in one boards don't bother with one, so this hasn't had much focus, but one feature that has not yet been implemented is support for swamping protection when TX and RX are close together, done by disabling the LNA when the signal is clipping. Currently the LNA is always on.

Another open issue is that the latest ACCST TX firmware / hardware from FrSky seems to introduce changes that break binding with the FrSky X SPI code: betaflight/betaflight#8603 I cannot do much about this for now as I do not have access to hardware that can run this firmware.

PR #70, opened 2019-10-18
Clean up pinio configuration for some unified targets

Docteh:
Couple things of note
- [x] Some targets specify `BOXARM` for use in `pinio_box` in their `config.c`, is this always 0?I thought this was a bad idea, and then I realized that in SPEEDYBEEF4 they are using it to turn off bluetooth when the pilot arms.
`src/main/msp/msp_box.c` says to not reuse box ids, so the use of zero is safe.

- [ ] EXUA-EXF722DUAL.config no `config.c` in betaflight
- [ ] VGOODRCF4 no `config.c` in betaflight, but [target.h](https://github.com/betaflight/betaflight/blob/c37a7c91a24d2828e0824225a52851bd0cfa40a6/src/main/target/VGOODRCF4/target.h#L128-L132) does say its for VTX power, and camera switcher
- [ ] TRANSTECF411 and TRANSTECF405 may have off by one error. Taking the time to say that `PINIO 2` is inverted, but there is no `PINIO 2`
- [ ] `IFRC-IFF7_TWIN_G.config:set motor_pwm_protocol = DSHOT1200` should ask someone about that one, I think we threw out DSHOT1200 for 4.1.0

I'm assuming that for `pinio_config` that 255 is what we want for something unassigned, and stomped out an instance each of `-1` and `0`


etracer65:
There's nothing wrong with using 0 to tie the `pinio` to the arming mode. Not sure I'd logically want to use that for something like a VTX power switch, but I guess you could. But make perfect sense for disabling bluetooth on the SpeedyBee boards.

Also, using mode 39 (VTX PIT MODE) for a VTX power control is a terribly bad idea. Basically what happens is that if the VTX is configured for SA or Tramp then whenever the VTX is powered it's also forced into pit mode. Should use mode 40 (USER1).

Docteh:
Looks like CLRACINGF_ is getting changed with  #71 and #72

But that still leaves
HENA-TALONF7FUSION.config:set pinio_box = 39,255,255,255
HENA-TALONF7V2.config:set pinio_box = 39,255,255,255
KLEE-SYNERGYF4.config:set pinio_box = 39,255,255,255
TMTR-TMOTORF7.config:set pinio_box = 39,255,255,255

For legacy targets SYNERGYF4 and the CLRACINGF4/F7 are using box 39

On one hand it seems like the manufacturers specified box 39, but maybe they misunderstood how vtxes work?

etracer65:
Yes, using 39 is a mistake- even in the legacy targets. It’s been a support with people having their VTX stuck in pit mode.

Docteh:
I submitted https://github.com/betaflight/betaflight/pull/9062, so we can consistently stamp out the  use of BOX 39 and PINIO, also updated the targets here.

For DSHOT1200 AIRB-XILOF4.config EXUA-EXF722DUAL.config IFRC-IFF7_TWIN_G.config

configs/default/HENA-TALONF7FUSION.config
```diff
@@ -80,7 +80,7 @@ serial 0 2048 115200 57600 0 115200
 serial 2 64 115200 57600 0 115200
 
 # aux
-aux 0 39 0 900 2100 0 0
+aux 0 40 0 900 2100 0 0
```
mikeller: I agree that changing the mode controlling the VTX power pin is something that _should_ be changed, as the existing setup _can_ lead to problems if users use a VTX that is controlled through Betaflight.
However, we need to make sure that the manufacturers of the hardware in question are in on this, and **change their documentation to reflect this change**. If not, we are changing something that will not work for _some_ users of this feature to something that will not work for _all_ users of this feature.
I think @bnn1044 is representing the manufacturer for this board?

configs/default/KLEE-SYNERGYF4.config
```diff
@@ -127,7 +127,7 @@ set pid_process_denom = 1
 set system_hse_mhz = 8
 set max7456_spi_bus = 3
 set dashboard_i2c_bus = 2
-set pinio_box = 39,255,255,255
+set pinio_box = 40,255,255,255
```
mikeller: @lexiotech: This will require the documentation that you provide for this board to be updated. Can you please confirm that you'll be doing this?

configs/default/TMTR-TMOTORF7.config
```diff
@@ -104,4 +104,4 @@ set dashboard_i2c_bus = 2
 set flash_spi_bus = 3
 set gyro_1_bustype = SPI
 set gyro_1_spibus = 1
-set pinio_box = 39,-1,-1,-1
+set pinio_box = 40,255,255,255
```
mikeller: @DieHertz: This will require the documentation that TMotor provides for this board to be updated. Can you please confirm that you'll be looking into this?

Docteh:
@mikeller I think I have what I want, 3 manufacturer specific PRs #75 #76 #77, and one PR with the general changes remaining here. Wording on those needs work.

Any thoughts on DSHOT1200? I think we removed it from betaflight, but not sure if that was before or after 4.1.0 released. Still mentioned in some targets.

etracer65:
Yes, we need to fix the DSHOT1200 references. No reason for the target to set any ESC protocol.

PR #73, opened 2019-10-19
Add OMNINXT7 Unified Target

RipperDrone:
assumed the MPU is STMF722, did find contradictory hints, though. Tried to flash 745 first, froze. 7X2 flashed successfully, therefore assuming it's correct. FC behaves normally after flashing, too.


jflyper:
Yes. It's F722RE.

RipperDrone:
Thank you for clarifying @jflyper - flyingfolk seems wrong then ;-): http://flyingfolk.com/F7-OmniNXT-Flugsteuerung-Flight-Controller-32khz-Dual-damped-Gyro

PR #76, opened 2019-10-19
HENA - Move power switch to be controlled by USER1 (40) instead of Pit mode (39)

Docteh:
This change will require a change in documentation. 

Assigning VTX Power switch to USER1 instead of VTX Pit mode allows the use of both modes without confusion.

### DIFF
--- a/configs/default/HENA-TALONF7FUSION.config
+++ b/configs/default/HENA-TALONF7FUSION.config
@@ -80,7 +80,7 @@ serial 0 2048 115200 57600 0 115200
 serial 2 64 115200 57600 0 115200
 
 # aux
-aux 0 39 0 900 2100 0 0
+aux 0 40 0 900 2100 0 0
 
 # master
 set gyro_to_use = BOTH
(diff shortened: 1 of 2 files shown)


mikeller:
Also, @bnn1044, should we put you down as the maintainer of HeliNation's targets?

bnn1044:
Yes. Please put me on CLRACING and pyrodrone. 

Thanks 

Cheng Lin
203-243-5683

mikeller:
@bnn1044: We don't have Unified Targets or a manufacturer for pyrodrone yet - the question was about HeliNation, for their Talon.. targets.

bnn1044:
@mikeller  I will add pyrodone later.  thanks. 
yes put me down for helination and CLRACING. Thanks

mikeller:
@bnn1044: See #78. This still leaves the open point of getting HeliNation to update the manual for their TalonF7 boards to show the 'VTX power switch' being assigned to the 'USER1' mode instead of 'PIT_MODE', as this makes the real pit mode unusable for users with controllable VTX.
Can you please ask them to do this, and then confirm?

bnn1044:
@mikeller I will let them know. thanks

PR #79, opened 2019-10-20
Update AlienFlight F4 unified configuration

MJ666:


configs/default/AFNG-ALIENFLIGHTF4.config
```diff
@@ -1,63 +1,60 @@
-# Betaflight / STM32F405 (S405) 4.0.0 Mar  2 2019 / 07:01:01 (29db27584) MSP API: 1.41
+# Betaflight / STM32F405 (S405) 4.1.0 Oct 16 2019 / 11:57:16 (c37a7c91a) MSP API: 1.42
+
+defaults nosave
```
etracer65: This should not be in the unified config.

MJ666:
@etracer65 fixed (also for other AFNG targets).

configs/default/AFNG-ALIENFLIGHTF4.config
```diff
@@ -84,71 +79,69 @@ timer C08 AF3
...
+set beeper_od = OFF
+set sdcard_detect_inverted = ON
+set sdcard_mode = SPI
+set sdcard_spi_bus = 2
+set flash_spi_bus = 2
+set gyro_1_bustype = SPI
+set gyro_1_spibus = 1
+set gyro_1_sensor_align = CW270
+set gyro_1_align_yaw = 2700
+set i2c1_pullup = ON
```
etracer65: Looks like the defaults for the current scale/offset defined in the target are not included.
```
#define CURRENT_METER_OFFSET_DEFAULT 2500                      // ACS712/714-30A - 0A = 2.5V
#define CURRENT_METER_SCALE_DEFAULT -667                       // ACS712/714-30A - 66.666 mV/A inverted mode
```

MJ666: We never finished development of Version 2 variants with the current sensor as we had planed and prototyped for a while. We progressed to the STM32F7. So this settings are obsolete.

PR #86, opened 2019-10-22
fix adc/rename/burst to AUTO

x4FF3:
Hi,

rename to correct MID, add missing adc entry and change dshot_burst to AUTO.

KR

### DIFF
--- a/configs/default/AIRB-XILOF4.config
+++ b/configs/default/AIRB-XILOF4.config
@@ -1,4 +1,5 @@
 # Betaflight / STM32F405 (S405) 4.1.0 May  6 2019 / 16:52:56 (011aeb66a) MSP API: 1.42
+# manufacturer_id: AIRB   board_name: XILOF4   custom defaults: NO
 
 board_name XILOF4
 manufacturer_id AIRB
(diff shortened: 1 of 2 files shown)


configs/default/AIRB-XILOF4.config
```diff
@@ -83,8 +86,8 @@ serial 1 1024 115200 57600 0 115200
 
 # master
 set blackbox_device = SPIFLASH
-set dshot_burst = ON
-set motor_pwm_protocol = DSHOT600
+set dshot_burst = AUTO
```
mikeller: Is this target designed for burst Dshot or is it not?

configs/default/LMNR-LUXMINIF7.config
```diff
@@ -1,8 +1,8 @@
-# Betaflight / STM32F7X2 (S7X2) 4.1.0 Sep 25 2019 / 14:09:57 (08e8afa09) MSP API: 1.42
-# manufacturer_id: AIRB   board_name: LUXMINIF7   custom defaults: NO
+# Betaflight / STM32F7X2 (S7X2) 4.1.0 Oct  6 2019 / 06:48:40 (c6452a55c) MSP API: 1.42
+# manufacturer_id: LMNR   board_name: LUXMINIF7   custom defaults: NO
 
 board_name LUXMINIF7
-manufacturer_id AIRB
+manufacturer_id LMNR
```
mikeller: @x4FF3: Should we put you down as the target maintainer for Lumenier as well?

x4FF3: yes please.

PR #101, opened 2019-11-01
Add IFRC-IFF7_TWIN_G_D

### LINKED ISSUE #98: Custom config needed for a fc with different gyro alignment?
I got a new FC from flight that took some custom config:

```# version
# Betaflight / STM32F7X2 (S7X2) 4.1.0 Oct 16 2019 / 11:58:45 (c37a7c91a) MSP API: 1.42
# manufacturer_id: EXUA   board_name: EXF722DUAL   custom defaults: YES
```

iFlight f7 twin-g for DJI air unit. Not the same as the other F7 twin-g stacks. 

When I tried to fly, it freaked out in a cyclical pattern each time because—as I found out via trial and error—one gyro was 90º off. I don't think the same custom defaults can apply for these two stacks (I have both).

image

I think iFlight needs to add a custom config for the board with different sensor alignment. I could be totally wrong here. This is my first issue to unified configs.

johnelliott:
Add iflight SucceX-D Mini F7

Adds support for the [iFlight SucceX-D Mini F7 board](https://www.iflight-rc.com/index.php?route=product/product&path=86_87&product_id=1060) which has different gyro alignment than the non-D board.

Fixes #98

@Linjieqiang here is my proposal to fix the issue with different hardware. Would you please confirm that this is different hardware?

### DIFF
--- a/configs/default/IFRC-IFF7_TWIN_G_D.config
+++ b/configs/default/IFRC-IFF7_TWIN_G_D.config
@@ -1,0 +1,136 @@
+# Betaflight / STM32F7X2 (S7X2) 4.0.3 Jun  1 2019 / 11:59:57 (094cfc956) MSP API: 1.41
+
+board_name IFF7_TWIN_G_D
+manufacturer_id IFRC
+
+# resources
+resource BEEPER 1 C15
+resource MOTOR 1 C08
+resource MOTOR 2 C06
+resource MOTOR 3 C09
+resource MOTOR 4 C07
+resource MOTOR 5 B06
+resource MOTOR 6 B07
+resource MOTOR 7 B01
+resource MOTOR 8 B00
+resource PPM 1 A03
+resource LED_STRIP 1 A01
+resource SERIAL_TX 1 A09
+resource SERIAL_TX 2 A02
+resource SERIAL_TX 3 B10
+resource SERIAL_TX 4 C10
+resource SERIAL_TX 5 C12
+resource SERIAL_RX 1 A10
+resource SERIAL_RX 2 A03
+resource SERIAL_RX 3 B11
+resource SERIAL_RX 4 C11
+resource SERIAL_RX 5 D02
+resource I2C_SCL 2 B10
+resource I2C_SDA 2 B11
+resource LED 1 C04
+resource SPI_SCK 1 A05
+resource SPI_SCK 2 B13
+resource SPI_SCK 3 B03
+resource SPI_MISO 1 A06
+resource SPI_MISO 2 B14
+resource SPI_MISO 3 B04
+resource SPI_MOSI 1 A07
+resource SPI_MOSI 2 B15
+resource SPI_MOSI 3 B05
+resource CAMERA_CONTROL 1 A00
+resource ADC_BATT 1 C01
+resource ADC_RSSI 1 C00
+resource ADC_CURR 1 C02
+resource PINIO 1 C13
+resource PINIO 2 C14
+resource PINIO 3 B08
+resource FLASH_CS 1 B09
+resource OSD_CS 1 B12
+resource GYRO_EXTI 1 A08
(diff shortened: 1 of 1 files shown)


configs/default/IFRC-IFF7_TWIN_G_D.config
```diff
@@ -0,0 +1,136 @@
...
+set beeper_inversion = ON
+set beeper_od = OFF
+set max7456_spi_bus = 2
+set dashboard_i2c_bus = 2
+set pinio_config = 129,1,1,1
+set pinio_box = 0,255,255,255
+set flash_spi_bus = 3
+set gyro_1_bustype = SPI
+set gyro_1_spibus = 1
+set gyro_1_sensor_align = CW270
```
johnelliott: gyro_x_sensor_align are the only departures from `configs/default/IFRC-IFF7_TWIN_G.config` L134 and L136

mikeller:
@johnelliott: Can you specify what other parts of the target configuration you have tested with your hardware? There might be other pins / UARTs / ... that are different to what is defined in IFRC-IFF7_TWIN_G.config.

johnelliott:
Hi @mikeller, I have this on a line-of-sight standard 4-rotor quad. I'm only using T1 for Frsky fport. I have not noticed other differences.

RipperDrone:
@johnelliott I tested Unified Target on TwinG-D F7 v3 today, and it seems timers are mixed up: Neither on cam nor on LED pin was I able to operate the camera OSD menu. Hardware OSD joystick is working ok, so cam control hardware seems ok...

RipperDrone:
Plus: Booting your UT configuration takes ~10 secs - maybe an indication of timer conflicts as well...?

johnelliott:
Hi @RipperDrone. Like I told Mike, I have only tested basic flight on a quad with just the stack + motors:
> Hi @mikeller, I have this on a line-of-sight standard 4-rotor quad. I'm only using T1 for Frsky fport. I have not noticed other differences.

If you've found more differences I'd say make a PR for them. I've put most of my quad stuff away for the season, so if it works for you we might as well just use your changes.

RipperDrone:
@johnelliott wish it was that easy for me - how should I proceed: Currently, I have no idea which timers / resources to re-assign to make it work again. Can you give me any pointers or refer to someone who knows better than me? iFlight seem to have adopted your Unified Target settings without checking, the timer/dma snippets are identical... :-O What should I do now?... :-/

RipperDrone:
Should I give the TwinG config a try or is it definitely required to use the TwinG_D version? It seems they have different timer assignments...

johnelliott:
@RipperDrone Yes, I see the iFlight website is now linking to the config committed here 😞. Maybe they thought their problem had gone away.

One thing I considered before was taking the config iFlight posted previously on their website for this board, checking that it is in fact for the right board and not them re-posting the config for the non-D version, and then taking the correct timer assignments form there and applying those to the config that worked with version 4.1. That appears to be messed up now. I _may_ have saved the old config on my computer, let me check for this.

If we don't have the old data then @Linjieqiang may be the best person to help with this. They have the responsibility for making this correct.

Docteh:
10 second delay sounds familiar, is the config looking for a MAG?

RipperDrone:
@Docteh @johnelliott Problem resolved! :-)

- looking for a mag was 5 secs out of the 10. SucceX Twin-G D F7 Unified Target defaults do not disable baro and mag
-  balance 5 secs boot delay seem normal on 2 gyro setups

RipperDrone:
Re OSD cam control function not working, this was a nasty one: Caddx for some reason decided to swap the labels on their Ratel camera: OSD pin is labelled GND and vice versa! After swapping the pins, cam control worked fine, some tuning of internal_resistance (270 Ohm) and 'down' resistance value (0 --> 15 kOhm) in CLI included :-) All good now!

PR #102, opened 2019-11-05
Fix gyro detection on specific targets

Asizon:
Since gyro detection for Dualgyro setups is improved, Dual gyro boards with only one gyro connected have been affected. When it is going to initialize configured by default in ```gyroauto```, it searches for gyroscopes in the spi port of gyro 2, since it does not have gyro installed, the search for this is delayed and a very long initialization is generated. I have disabled this search for these target. Also @airbotfpv should include in its manuals how to activate Gyro 2 for those who want it.
This fix #9144 https://github.com/betaflight/betaflight/issues/9144

### DIFF
--- a/configs/default/AIRB-AIRBOTF7.config
+++ b/configs/default/AIRB-AIRBOTF7.config
@@ -88,4 +88,6 @@ set gyro_1_bustype = SPI
 set gyro_1_spibus = 3
 set gyro_1_sensor_align = CW90
 set gyro_1_align_yaw = 900
+set gyro_2_bustype = NONE
 set gyro_2_spibus = 1
+
--- a/configs/default/AIRB-AIRBOTF7.config
+++ b/configs/default/AIRB-AIRBOTF7.config
@@ -90,4 +90,3 @@ set gyro_1_sensor_align = CW90
 set gyro_1_align_yaw = 900
 set gyro_2_bustype = NONE
 set gyro_2_spibus = 1
-
--- a/configs/default/AIRB-OMNIBUSF4FW.config
+++ b/configs/default/AIRB-OMNIBUSF4FW.config
@@ -127,5 +127,6 @@ set flash_spi_bus = 2
 set gyro_1_bustype = SPI
 set gyro_1_spibus = 3
 set gyro_1_sensor_align = CW180
+set gyro_2_bustype = NONE
 set gyro_2_spibus = 1
 set gyro_2_sensor_align = CW0FLIP
--- a/configs/default/AIRB-OMNIBUSF4NANOV7.config
+++ b/configs/default/AIRB-OMNIBUSF4NANOV7.config
@@ -94,5 +94,6 @@ set flash_spi_bus = 1
 set gyro_1_bustype = SPI
 set gyro_1_spibus = 3
 set gyro_1_sensor_align = CW0
+set gyro_2_bustype = NONE
 set gyro_2_spibus = 1
 set gyro_2_sensor_align = CW0
--- a/configs/default/AIRB-OMNIBUSF4V6.config
+++ b/configs/default/AIRB-OMNIBUSF4V6.config
@@ -125,5 +125,6 @@ set flash_spi_bus = 2
 set gyro_1_bustype = SPI
 set gyro_1_spibus = 1
 set gyro_1_sensor_align = CW180
+set gyro_2_bustype = NONE
 set gyro_2_spibus = 1
 set gyro_2_sensor_align = CW0
--- a/configs/default/AIRB-OMNIBUSF7NANOV7.config
+++ b/configs/default/AIRB-OMNIBUSF7NANOV7.config
@@ -93,4 +93,5 @@ set flash_spi_bus = 1
 set gyro_1_bustype = SPI
 set gyro_1_spibus = 3
 set gyro_1_sensor_align = CW0
+set gyro_2_bustype = NONE
 set gyro_2_spibus = 3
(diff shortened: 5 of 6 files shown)


mikeller:
I am strongly against this - this seems to be trading a minor inconvenience (~ 5 seconds delayed boot time while detection of the second gyro is attempted) for a problem that be seen as a breaking change by a lot of users (second gyro not found in default configuration for all owners of boards with a second gyro).
@x4FF3: If this change goes ahead there will be backlash / support issues created by a large number of users owning any of these boards with 2 gyros installed - are you and Airbot willing to take on answering these support calls?

Asizon:
I think it is a very large delay, it is not normal that much waiting in initialization, really that the delay is quite significant. I'm worried about having to wait in the races to initialize to make sure it's right before I start flying(example).Also the affected users will be practically none, so it is seen nobody uses these boards with external gyro installed

mikeller:
@Asizon:

> Also the affected users will be practically none, so it is seen nobody uses these boards with external gyro installed

Did you get this data from Airbot, or can you state where you got the information from?

@x4FF3: Can you please confirm @Asizon's statement that none of these boards are used with a second gyro installed? If so we can publicly state that a second gyro on these boards is not supported by Airbot and close any issues arising from this.

Asizon:
Also, when the unified target is 100% implemented on users' boards, I think we will start receiving initialization delay issues.

x4FF3:
@Asizon @mikeller we will test this further, and discuss it. based on those tests we will come back to this PR and decide how to proceed.

Asizon:
@mikeller I have always supported to use these controllers with an external gyro box installed, but it seems that nobody is encouraged to install this :( . Anyway, if this affects many users and damages Betaflight, we could forget this PR, write a note on the wiki of how to improve the boot, and set this in the following flight controller models Target

Docteh:
@Asizon how about a `Slow Startup` page under troubleshooting?

image

Asizon:
@Docteh This would be great, we will wait for the tests and proposal of @x4FF3 and @airbot_fpv

Asizon:
@mikeller can you put label to this PR? We are waiting for Airbot answer and tests

mikeller:
@Asizon: I think the fact that Airbot did not want to comment on this for over a month is evidence enough that this should be closed.

x4FF3:
Hi,
i am sorry, couldnt find to test this properly. But the short tests i could do, let me come to the conclusion, that an entry in the wiki would be the way to go. @Asizon @Docteh

Asizon:
What do you think? Need to add boards examples experimenting this issue, some anything?
https://github.com/betaflight/betaflight/wiki/slow-startup

mikeller:
@Asizon: Renamed the page to make the title accurate: https://github.com/betaflight/betaflight/wiki/Slow-Startup-on-Dual-Gyro-Boards-with-Only-One-Gyro-Installed

mikeller:
@x4FF3: Does the fact that you have reopened this mean that it should be merged, and we should tell every user of a dual gyro Airbot board to return it for a refund because the hardware they paid for is not supported any more?

AirbotFPV:
The 2nd gyro is optional for most of the Airbot FCs when the customer buys the board, they have been clearly told that it is one gyro or two gyros.

I think for this moment the wiki is fine, and I will remind the user about this too. 

On the other hand, we can remove the 2nd gyro support for the NONE-2nd gyro pre-mounted board. But based on our gyro mounting structure, I think there is a possibility that the user removes the gyro box and uses only the onboard Gyro. (then it means the wiki page still needed).

AirbotFPV:
For example, for Airbot F7,  it is not designed for 2 gyros, 
https://store.myairbot.com/airbotf7.html
But we had one FPC plug there that you can connect one ext. gyro box(to make the FC as 2 gyro version), or put the Frsky Receiver on top. 

In this regard, we asked the two gyro support on the Airbot F7 code. but most users will only use one gyro. 

another example will be:
https://store.myairbot.com/omninxtf7.html
we sell the board as a combo, that it always had two gyros(onboard + the one in the gyro box). 
So we have dual gyro support in the target file too. 
But it is possible if the customer removes the gyro box to lower the profile(then use only the onboard MPU6000.

mikeller:
@AirbotFPV: You'll have to sort this out with @x4FF3, as he is registered as the maintainer of the Airbot targets with us, and he has re-opened this pull request. So if he does not close this again then this will be merged and we will direct users opening issues because their second gyro stopped working that the second gyro on Airbot hardware is not supported.

Asizon:
Ok, thanks for the correction @mikeller . As far as I can understand, @AirbotFPV has decided that it is not necessary to continue with this PR, for now the wiki document is our solution. I have plubished it in Troubleshooting chapter.
slowstartup

Asizon:
Also I believe that gyro auto detection can be improved in the future to solve these problems :)

AirbotFPV:
@mikeller  Got it, I will remind him when he is back from holiday:), by the way, do you think the Gyro Detection could be changed in the future? 

@Asizon  Thanks for the document.

mikeller:
@Asizon: I think you will find that most developers will prefer to spend their time on fixing things that are not only cosmetic issues (most users won't be ready to fly by the time gyro initialisation has finished even on dual gyro boards with only one gyro installed). So if you think this is a problem then it's probably best if you open a pull request that fixes it.

@AirbotFPV: There are certainly options - the gyro probing could be rewritten to probe multiple gyros at once, or to probe first and second gyro in parallel. Or the whole initialisation could be rewritten to initialise all peripherals in parallel. Let's wait for @Asizon's pull request.

Asizon:
Understood @mikeller , I would love to have the necessary knowledge to do this, even so, I also think that this is only cosmetic and the dear time should be used in other things, just talked about a distant future.

PR #103, opened 2019-11-08
Add missing -RX_PARALLEL_PWM

Asizon:
Added missing ```-RX_PARALLEL_PWM``` to CrazybeeF4FS config

### DIFF
--- a/configs/default/HAMO-CRAZYBEEF4FS.config
+++ b/configs/default/HAMO-CRAZYBEEF4FS.config
@@ -79,6 +79,7 @@ dma pin A10 0
 # feature
 feature OSD
 feature RX_SPI
+feature -RX_PARALLEL_PWM
 
 # master
 set dshot_burst = AUTO


etracer65:
This is unnecessary. Just enabling `feature RX_SPI` is sufficient. Enabling any of the other receiver types will cause feature `RX_PARALLEL_PWM` to be disabled.

Asizon:
@etracer65 thanks for the point, I had that doubt, but after enabling SPI Rx, the ```diff all bare``` command answered me with ```feature -RX_PARALLEL_PWM```

mikeller:
@etracer65: I agree with @Asizon on this one: if a different RX type is configured in a Unified Target configuration, then there will be an artefact in a clean `diff` if `RX_PARALLEL_PWM`  is not disabled:

```
# feature
feature -RX_PARALLEL_PWM
```

We should potentially look into dropping `RX_PARALLEL_PWM` as the default RX provider, as this seems to be an anachronism. The problem there is that with the way the config validation works, `RX_SERIAL` (which is the most commonly used RX provider) will be disabled if no UART is configured for it.

image

PR #105, opened 2019-11-13
Initial MAMBAF405US Config

kc10kevin:
MAMBAF405US Config File for Betaflight V4.1.0.

### DIFF
--- a/configs/default/DIAT-MAMBAF405US.config
+++ b/configs/default/DIAT-MAMBAF405US.config
@@ -1,0 +1,118 @@
+# Betaflight / STM32F405 (S405) 4.1.0 Oct 16 2019 / 11:57:16 (c37a7c91a) MSP API: 1.42
+# manufacturer_id: DIAT   board_name: MAMBAF405US   custom defaults: NO
+
+board_name MAMBAF405US
+manufacturer_id DIAT
+
+# I/O
+resource LED 1 C15
+resource LED 2 C14
+resource BEEPER 1 C13
+set beeper_inversion = ON
+set beeper_od = OFF
+
+# ACC/GYRO
+resource GYRO_EXTI 1 C04
+resource GYRO_CS 1 A04
+set gyro_1_bustype = SPI
+set gyro_1_spibus = 1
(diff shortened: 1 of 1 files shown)


configs/default/DIAT-MAMBAF405US.config
```diff
@@ -0,0 +1,118 @@
...
+resource PINIO 1 B00
+
+# master
+set mag_hardware = NONE
+set baro_hardware = NONE
+set serialrx_provider = SBUS
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
+set pinio_box = 0,255,255,255
```
mikeller: Is this to disable a Bluetooth transceiver when armed?

Is it possible to get an engineering sample of this board?

mikeller:
Also, @kc10kevin, are you the maintainer of the Diatone target configurations, and should I give you approval rights for them?

configs/default/DIAT-MAMBAF405US.config
kc10kevin: @mikeller, TBH, I'm not 100% sure on that config code.   I'm still trying to figure out that Config setting.    I'm going off of what the "experts" did when adding that to the Config files to accommodate the pinio from the Legacy target to the Unified Targets.   "Legacy" target of the board would have the config.c for the bluetooth:

https://github.com/kc10kevin/betaflight/blob/master_410/src/main/target/MAMBAF405US/config.c

Pretty much the same as the MAMBAF722 board (couple of pin differences), so used the same pinio settings in the Config file since they are not exported.

Boards are in prototype status right now.   Testing complete, going to production now.   If you want one, I'll be happy to contact Diatone and have one sent to you.   I'm sure they will gladly send one to you.   You've got my e-mail, so send me your address and I'll send it to them.

kc10kevin:
@mikeller, Yes I am.   I maintain all of the Diatone target configurations for them.   MAMBA and FURY line of products.   Please give me approval rights.  Much appreciated!!!

PR #129, opened 2019-12-05
Added target EXUAVF4PRO.

mikeller:


etracer65:
Also looks to need
```
feature TELEMETRY
```

mikeller:
Removed non-on-board UART pre-allocations.

PR #130, opened 2019-12-05
Added target AG3XF7.

mikeller:


mikeller:
I am actually not a big fan of manufacturers pre-allocating UARTs just for the sake of it - in my experience this turns out to be not what the user wants at least as many times as not. Removed.

PR #144, opened 2019-12-26
add KAKUTEF7HDV target for holybro F745 FC

jamming:
KAKUTEF7HDV suitable for dji air unit. It doesn't have a max7456 osd chip. The PID Loop is unstable  if enabling the max7456.

### DIFF
--- a/configs/default/HBRO-KAKUTEF7HDV.config
+++ b/configs/default/HBRO-KAKUTEF7HDV.config
@@ -1,0 +1,99 @@
+# Betaflight / STM32F745 (S745) 4.0.0 Apr  3 2019 / 14:32:23 (22b9f3453) MSP API: 1.41
+
+board_name KAKUTEF7HDV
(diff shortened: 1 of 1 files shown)


Asizon:
LOL, timers for this target config are completly missing???? @jamming

mikeller:
@Asizon: Not sure where you are looking, but I can see the timers: https://github.com/betaflight/unified-targets/pull/144/files#diff-c05ec80e13b6a54e1edef1daab5f168dR50

Asizon:
Sorry, I didn't know that representation of timers.

mikeller:
Good point actually, that is the legacy format for timers that we should be dropping.

@jamming: Can you please change this target to use the AF based format for the timer configuration? The format that you are using will be dropped sooner or later.

mikeller:
@jamming: betaflight/betaflight#9484 means this target will be broken in 4.2 unless it is fixed.

PR #148, opened 2019-12-29
Merge pull request #148 from Linjieqiang/fix_ifrc_iff4_twin_g

Linjieqiang:


configs/default/IFRC-IFF4_TWIN_G.config
```diff
@@ -78,6 +79,8 @@ dma pin B06 0
 # pin B06: DMA1 Stream 0 Channel 2
 dma pin A00 0
 # pin A00: DMA1 Stream 5 Channel 3
+dma pin A00 0
```
mikeller: Looks like this is wrong (it is a double up of the line above). Should this be A01?

Linjieqiang: Oh. My wrong. I will fix it soon.

PR #152, opened 2019-12-30
Create BEFH-BETAFPVF405

betafpv-engineer:


configs/default/BEFH-BETAFPVFF405.config
```diff
@@ -0,0 +1,137 @@
+# Betaflight / STM32F405 (S405) 4.1.2 Dec 22 2019 / 02:19:35 (89051e256) MSP API: 1.42
+
+board_name BETAFPVF405
```
mikeller: Looks like there is a spelling mistake in the file name: `BEFH-BETAFPVFF405.config`.

betafpv-engineer: Yes, I found it was my negligence and I re-uploaded it.

PR #154, opened 2019-12-30
set the default current scale for holybro F4/F7 FC

jamming:
The current meter scale of holybro F4/F7 FC or ESC is 275. Set it to default, because it's difficult for a user to calibrate without a dedicated device.


jamming:
On KakuteF7HDV, UART1 is used for telemetry port to communicate with DJI air unit. UART6 is used as RC receiver. These two UARTs are wired into  a 8 pins connector for the connection of DJI air unit. Configure it with a default setting to make it easier to use.

PR #155, opened 2020-01-03
add MTKS-MATEKF722PX.config

MATEKSYS:
no MAX7456 on this FC, but if no definitions for MAX7456,  Font manager will not display.

### DIFF
--- a/configs/default/MTKS-MATEKF722PX.config
+++ b/configs/default/MTKS-MATEKF722PX.config
@@ -1,0 +1,154 @@
+# Betaflight / STM32F7X2 (S7X2) 4.1.0 Sep  8 2019 / 01:19:23 (f331cb3dd) MSP API: 1.42
+
+board_name MATEKF722PX
+manufacturer_id MTKS
+
+# RESOURCE
+resource BEEPER 1 C13
+resource MOTOR 1 C08
+resource MOTOR 2 C09
+resource MOTOR 3 B04
+resource MOTOR 4 B05
+resource MOTOR 5 B00
+resource MOTOR 6 B01
+resource MOTOR 7 B10
+resource MOTOR 8 B11
+resource SERVO 1 B06
+resource SERVO 2 B07
+resource PPM 1 A03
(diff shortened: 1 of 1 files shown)


Docteh:
If there is no MAX7456 chip should `max7456_spi_bus` be set to `0`?
https://github.com/betaflight/betaflight/issues/9287

MATEKSYS:
> If there is no MAX7456 chip should `max7456_spi_bus` be set to `0`?
> [betaflight/betaflight#9287](https://github.com/betaflight/betaflight/issues/9287)
must define max7456 SPI and OSD_CS, otherwise no font manager

https://github.com/betaflight/betaflight/pull/9158

mikeller:
Currently font uploading is only supported for the MAX7456 hardware, so there is no need to display the font manager if there is no MAX7456.

If the hardware does not have a MAX7456 then `set max7456_spi_bus = 2` is wrong.

mikeller:
@MATEKSYS:

> FrskyOSD uses same font manager as MAX7456.
> When FrskyOSD is enabled on Uart. font will be uploaded to FrskyOSD.

No, it does not in the current version of Betaflight.

MATEKSYS:
@mikeller it works fine in BF4.2, just no font manager if no max7456 definition.

mikeller:
'BF4.2' is not something that is released or supported by Betaflight. If you want your hardware to work with non-Betaflight firmware then adding target configurations here is the wrong thing to do.

MATEKSYS:
https://github.com/betaflight/betaflight/pull/9127
I have seen FrskyOSD was supported.

mikeller:
@MATEKSYS:

- FrSky OSD will be supported by Betaflight after 4.2 has been released - there is no date set for this yet:

image

- even when Betaflight will support FrSky OSD in 4.2, FrSky is not paying us to add support for configuring their hardware to configurator, so we are not planning to do this - they will have to provide their own tooling to configure their hardware.

PR #157, opened 2020-01-08
Add IFF4_AIO none OSD version Config

Linjieqiang:
According to user feedback, we made two versions of the fc hardware board, one with OSD and one without OSD. 

However, we found that the firmware did not enable the automatic detection of the OSD chip, which resulted in the inability to perform flight unlocking without OSD.

The video is here: https://www.youtube.com/watch?v=_EEoW1pXgVU&feature=youtu.be 

So we need two separate configuration files.

### DIFF
--- a/configs/default/IFRC-IFF4_AIO_OSD.config
+++ b/configs/default/IFRC-IFF4_AIO_OSD.config
@@ -1,6 +1,6 @@
 # Betaflight / STM32F405 (S405) 4.0.0 Mar 14 2019 / 11:45:26 (360afd96d) MSP API: 1.41
 
-board_name IFF4_AIO 
+board_name IFF4_AIO_OSD 
 manufacturer_id IFRC
 
 # resources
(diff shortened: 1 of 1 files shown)


configs/default/IFRC-IFF4_AIO.config
```diff
@@ -81,7 +80,6 @@ feature RX_SERIAL
 feature TELEMETRY
 feature LED_STRIP
 feature DISPLAY
-feature OSD
```
mikeller: Not sure if this is the right way to do this - this will mean that existing users of the IFF4_AIO board will get this installed on a firmware upgrade, and then their OSD will stop working, and they will open a support case.
Probably a better idea to leave this as it is, and add a new target `IFRC-IFF4_AIO_DJI` or similar.

mikeller:
Hold off on this for now - betaflight/betaflight#9360 should make it possible to use the regular target for this. You can test this with the firmware files published in https://github.com/betaflight/betaflight/pull/9360#issuecomment-574615784.

Linjieqiang:
@mikeller Thanks. Let me have a try to test with it .

Linjieqiang:
Tested with the firmware files published in https://github.com/betaflight/betaflight/pull/9360#issuecomment-574615784 . Worked well. Thanks @mikeller

PR #158, opened 2020-01-13
Allow bidirectional dshot on IFF411_PRO and TWING and provide alterna…

joelucid:
…tive config for TRANSTECF411 to run bidir without led strip

The IFF boards have the traditional timer 1 conflict with f411 which we handle by using timer dshot on boards which support it. The TranstecF411 can't do timer dshot due to one motor pin not having dma support. Here I provide an alternative config which disables the led strip.


joelucid:
Build fails for the transtec due to naming issues. @mikeller please let me know how we can add this config.

fujin:
@joelucid the build fails as the filename is `TTRH_BIDIR-*` but the manufacturer ID is `TTRH`. Suggest changing it to that (`TTRH`) (and perhaps adjusting the target name if you want to indicate bidir capabilities, although we have to contend with API pagination/limits here?). 

manufacturer ID variations such as _BIDIR suffix are not currently possible.

joelucid:
@fujin - ok, I renamed it to TTRH-TRANSTECF411_RPMFILTER now. Probably makes it even easier to find.

configs/default/TTRH-TRANSTECF411_RPMFILTER
```diff
@@ -0,0 +1,86 @@
+# Betaflight / STM32F411 (S411) 4.1.0 Sep  5 2019 / 19:50:56 (2553a4a1f) MSP API: 1.42
+
+board_name TRANSTECF411
```
etracer65: I don't think that commenting out the `LED_STRIP` resource is a valid reason to create a different target variant (TTRH-TRANSTECF411_RPMFILTER). If we were going down that path we'd end up with all kinds of target variants for alternate functions on flight controllers. Alternatives like one that enables softserial on a LED pin; one that resources camera control to motor 6; and so on. This squarely falls into the manufacturer documentation responsibility.

joelucid: Ok - I've removed it. It just comes up a lot recently. I think I've seen like 5 people with that issue last week. Fortunately there are only few boards which we can't fix in a non conflicting way.

PR #162, opened 2020-01-17
Add HGLR-HGLRCF722.config

HGLRC-D:
微信图片_20200117114130


mikeller:
This is in the wrong repository - please open this in https://github.com/betaflight/unified-targets/configs/default.

MATEKSYS:
Pins definitions 100% copy from MATEKF722mini

mikeller:
@MATEKSYS: Is there any problem with these pin definitions that you would like to report?

PR #166, opened 2020-02-01
solved TIMER conflict with LEDStip and M1 / M4

Guidus93:
Solved TIMER conflict with LEDSTRIP and M1/M4 with BLHeliS ESC


Guidus93:
I should be allowed to push my target but git gave me 403 error and told me that im not allowed to push

mikeller:
@Guidus93: No, you will still have to go through a pull request to get changes into `master` - however you are listed as a maintainer for ZEEZ- targets, so you will be able to approve / merge them (but unfortunately only if you are not the one opening them, I have not quite figured out how to address this outside of using a second account to open the pull request).

PR #170, opened 2020-02-02
Added target FF_PIKOF4.

mikeller:


configs/default/FFPV-FF_PIKOF4.config
```diff
@@ -0,0 +1,75 @@
+# diff all bare
+
+# version
+# Betaflight / STM32F405 (S405) 4.2.0 Feb  2 2020 / 14:39:25 (30bf9e809f) MSP API: 1.43
+# manufacturer_id: AIRB   board_name: AG3XF4   custom defaults: NO
```
etracer65: This should be removed. Also some extra stuff at the top.

mikeller:
Fixed.

PR #195, opened 2020-02-02
Create TCMM-TCMMF7.config

DMAXYANG:
This target Based on target MATEKF722SE
I change ‘set gyro_1_sensor_align = CW180FLIP’ to 'set gyro_1_sensor_align = CW180'

### DIFF
--- a/configs/default/TCMM-TCMMF7.config
+++ b/configs/default/TCMM-TCMMF7.config
@@ -1,0 +1,146 @@
+# Betaflight / STM32F7X2 (S7X2) 4.1.2 Dec 22 2019 / 02:21:16 (89051e256) MSP API: 1.42
+
+board_name TCMMF7
+manufacturer_id TCMM
+
+# resources
+resource BEEPER 1 C13
+resource MOTOR 1 B04
+resource MOTOR 2 B05
+resource MOTOR 3 B00
+resource MOTOR 4 B01
+resource MOTOR 5 A15
+resource MOTOR 6 B03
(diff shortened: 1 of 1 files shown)


mikeller:
@DMAXYANG: Are you the maintainer for TCMM targets?

DMAXYANG:
Yes, I am. Thank you

DMAXYANG:
Increase default core temperature threshold
Because my FC and VTX are integrated together, FC core temperature is very high without cooling.
Can you add code to TCMMF7.config file? 
For example:
#ifdef TCMMF7
#define CORE_TEMPERATURE_WARNING_THRESHOLD 125

mikeller:
@DMAXYANG: This is not how it will work, as the Unified Target configurations are not source code that is compiled. But you can add the following to your target configuration:

```
set core_temp_alarm = <temperature>
```

Please note also that 125 is most likely way too high - most STM MCUs seem to develop reliability issues as soon as the reported core temperature goes above 85 C - if you set the alarm above this then your customers will experience in-air lock-ups and crashes without any prior warning - not sure this will be good for the reputation of your products.

DMAXYANG:
@mikeller Thank you for your advice.
I tested that the core temperature during the flight was only 45 C，I will increase the cooling area

mikeller:
@DMAXYANG:

> I tested that the core temperature during the flight was only 45 C，I will increase the cooling area

It's not uncommon to have overheating issues only when not in flight, due to the high airflow generated by the props.

PR #196, opened 2020-02-03
Sdcard mode update

TheIsotopes:
sdcard_mode was set to SDIO in https://github.com/betaflight/betaflight/pull/9436

this pr is supposed to fix that


mikeller:
@TheIsotopes: `sdcard_mode` being set to `sdio` was an oversight and is fixed in #9445.

TheIsotopes:
if it fixed this and merged I'll close it.

PR #222, opened 2020-03-21
Disabling the VTX switch on TALONF7V2 due to thermal problems caused by it.

mikeller:
Fixes betaflight/betaflight#9516.

@bnn1044: I think this needs some more investigation, but it looks like the hardware is defective.

### DIFF
--- a/configs/default/HENA-TALONF7V2.config
+++ b/configs/default/HENA-TALONF7V2.config
@@ -40,7 +40,9 @@ resource CAMERA_CONTROL 1 B03
 resource ADC_BATT 1 C02
 resource ADC_RSSI 1 C03
 resource ADC_CURR 1 C01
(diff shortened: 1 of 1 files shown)


bnn1044:
@mikeller I don't think there is any hardware issue. I will investigate it and let you know. 
Thanks.

  The F722 running at 85c is not over heat base on their manual. 
Operating temperatures
Ambient temperatures: –40 to +85 °C /–40 to +105 °C
Junction temperature: –40 to + 125 °C

mikeller:
@bnn1044: Thanks for investigating.
Unfortunately the STM documentation seems to be overly optimistic about the temperature range - according to user reports the MCUs start to get unreliable if they go much over 85 C, resulting in loss of control and crashes - that's why the default temperature warning is set for 85 C.

PR #224, opened 2020-03-21
Added MATEKF405AIO target.

IgnasD:
Added target for [Matek F405-AIO board](http://www.mateksys.com/?portfolio=f405-aio).

Used latest available version (4.1.1) of legacy MATEKF405 target as a hardware dump base. Applied dump on unified STM32F405 target of the same version. Corrected gyro alignment and amperage meter scale. Disabled baro and mag.

To better align with existing MATEKF405xxx targets, switched RX to serial SBUS, ESC protocol to DSHOT600, enabled OSD, telemetry and soft serial.

### DIFF
--- a/configs/default/MTKS-MATEKF405AIO.config
+++ b/configs/default/MTKS-MATEKF405AIO.config
@@ -1,0 +1,143 @@
+# Betaflight / STM32F405 (S405) 4.1.1 Nov 15 2019 / 12:54:53 (1e5e3d369) MSP API: 1.42
+# manufacturer_id: MTKS   board_name: MATEKF405AIO   custom defaults: NO
+
+board_name MATEKF405AIO
+manufacturer_id MTKS
+
+# resources
+resource BEEPER 1 C13
+resource MOTOR 1 C06
+resource MOTOR 2 C07
+resource MOTOR 3 C08
+resource MOTOR 4 C09
+resource MOTOR 5 A15
+resource MOTOR 6 A08
+resource MOTOR 7 B08
+resource PPM 1 A03
+resource PWM 1 A00
+resource PWM 2 A01
+resource PWM 3 A02
+resource LED_STRIP 1 B06
+resource SERIAL_TX 1 A09
+resource SERIAL_TX 2 A02
+resource SERIAL_TX 3 C10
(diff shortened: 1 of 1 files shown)


configs/default/MTKS-MATEKF405AIO.config
```diff
@@ -0,0 +1,144 @@
...
+# pin B06: DMA1 Stream 0 Channel 2
+dma pin A00 0
+# pin A00: DMA1 Stream 2 Channel 6
+dma pin A01 0
+# pin A01: DMA1 Stream 4 Channel 6
+
+# feature
+feature -RX_PARALLEL_PWM
+feature RX_SERIAL
+feature SOFTSERIAL
```
etracer65: Since there are no resources allocated for softserial rx/tx there's no point enabling the feature.

configs/default/MTKS-MATEKF405AIO.config
```diff
@@ -0,0 +1,144 @@
...
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set mag_hardware = NONE
+set baro_bustype = I2C
+set baro_i2c_device = 1
+set baro_hardware = NONE
+set serialrx_provider = SBUS
+set blackbox_device = SDCARD
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
mikeller: This will cause motors to spin up on power up with older ESCs that do not support Dshot and recognise Dshot frames as valid PWM input.
@MATEKSYS will have to confirm that they are willing to assume this risk and take responsibility if users get injured because of this.

configs/default/MTKS-MATEKF405AIO.config
IgnasD: I agree, but the idea was to align with existing MATEKF405xxx targets which do the same. See:
https://github.com/betaflight/unified-targets/blob/master/configs/default/MTKS-MATEKF405CTR.config#L112
https://github.com/betaflight/unified-targets/blob/master/configs/default/MTKS-MATEKF405MINI.config#L110
https://github.com/betaflight/unified-targets/blob/master/configs/default/MTKS-MATEKF405STD.config#L117

Anyway, will do as requested.

PR #226, opened 2020-03-24
Create FOXE-FOXEERF722V2.config

Johnsnow-foxeer:
I am Foxeer target owner from Foxeer, Shenzhen, we want to upgrade to F722 V2 firmware, in the V2 version, the 20602 Gyro is cancelled.


DieHertz:
Please add this guy as FOXEER target owner

PR #233, opened 2020-04-06
Fixed gyro alignment for AGXF7 board.

mikeller:
Fixes betaflight/betaflight#9462.

Adjusted default gyro orientation for the on-board gyro to match the motor ordering as documented in the manual supplied with the board.

@jflyper: Can you review please? Also, can you please check if the AG3Xf4 needs this as well? I don't have an engineering sample for that one.


configs/default/AIRB-AG3XF7.config
```diff
@@ -128,8 +128,9 @@ set dashboard_i2c_bus = 2
 set flash_spi_bus = 2
 set gyro_1_bustype = SPI
 set gyro_1_spibus = 1
-set gyro_1_sensor_align = CW0FLIP
+set gyro_1_sensor_align = CW270FLIP
```
x4FF3, line 131: Hi,

just got the board and tested it. CW90FLIP is correct.

x4FF3:
Hi,

just got the board out of storage and tested it. 270 is correct :)

KR

x4FF3:
> @x4FF3: You wouln't happen to have the F4 version of the board as well, so that you can test if this needs a change of gyro alignment as well?

nope unfortunately not. but by the board file, i can tell it has the same orientation.

PR #235, opened 2020-04-11
add new target GRAVITYF7

cvetaevvitaliy:
MCU: STM32F722RET6 216MHz 32bit
UART: 6
Baro: BMP280
Gyro: MPU6000
ESC telem: RX6
SBUS: RX1 - (inverter)
S.PORT: TX4
SmartAudio: TX3
CameraControl: PB1
Power in: 3-6S Lipo (12.6-25V)
BEC: 5V/3А, 9V/3A (9V or 5V) 
USER1 - PC2 pin
LED indicate: green (D4) power STM32 (3.3V), red (D3) status FC, red (VTX) BEC 9V
OSD: AT7456E

### DIFF
--- a/configs/default/FLMD-DRONIUSF7.config
+++ b/configs/default/FLMD-DRONIUSF7.config
@@ -1,0 +1,111 @@
+# Betaflight / STM32F7X2 (S7X2) 4.1.5 Mar 16 2020 / 05:21:46 (d4e74e39c) MSP API: 1.42
+
+board_name DRONIUSF7
+manufacturer_id FLMD
+
+# resources
+resource BEEPER 1 C13
+resource MOTOR 1 A09
+resource MOTOR 2 A08
+resource MOTOR 3 C08
+resource MOTOR 4 C09
+resource PPM 1 B07
+resource LED_STRIP 1 A15
+resource SERIAL_TX 1 B06
+resource SERIAL_TX 2 A02
+resource SERIAL_TX 3 C10
+resource SERIAL_TX 4 A00
+resource SERIAL_TX 5 C12
+resource SERIAL_TX 6 C06
+resource SERIAL_RX 1 B07
+resource SERIAL_RX 2 A03
+resource SERIAL_RX 3 C11
+resource SERIAL_RX 4 A01
+resource SERIAL_RX 5 D02
+resource SERIAL_RX 6 C07
+resource I2C_SCL 1 B08
+resource I2C_SDA 1 B09
+resource LED 1 C14
+resource SPI_SCK 1 A05
+resource SPI_SCK 2 B13
+resource SPI_SCK 3 B03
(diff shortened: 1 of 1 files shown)


configs/default/FLMD-DRONIUSF7.config
```diff
@@ -0,0 +1,111 @@
+# Betaflight / STM32F7X2 (S7X2) 4.1.5 Mar 16 2020 / 05:21:46 (d4e74e39c) MSP API: 1.42
+
+board_name DRONIUSF7
```
etracer65, line 3: Why is the board name not "GRAVITYF7"?  Also the file is named in correctly.

configs/default/FLMD-DRONIUSF7.config
```diff
@@ -0,0 +1,111 @@
...
+feature OSD
+
+# serial
+serial 0 64 115200 57600 0 115200
+serial 2 2048 115200 57600 0 115200
+
+# master
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set mag_hardware = NONE
```
etracer65, line 91: There's no reason to disable the magnetometer. All this does is make it hard for someone if that want to add an external magnetometer later.

configs/default/FLMD-DRONIUSF7.config
```diff
@@ -0,0 +1,111 @@
...
+set baro_bustype = I2C
+set baro_i2c_device = 1
+set baro_i2c_address = 119
```
etracer65, line 94: You shouldn't need to specify the I2C address - the autodetect will deal with this.

configs/default/FLMD-DRONIUSF7.config
```diff
@@ -0,0 +1,111 @@
...
+set serialrx_provider = SBUS
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set current_meter = ADC
+set battery_meter = ADC
+set ibata_scale = 88
```
etracer65, line 100: This is an extremely unusually low value. Are you sure this is correct? Does the board have a built-in current sensor? If not then this shouldn't be set.

configs/default/FLMD-DRONIUSF7.config
cvetaevvitaliy: nop! This latest config in repo, I del this config

configs/default/FLMD-DRONIUSF7.config
cvetaevvitaliy: nop - scale - 182

configs/default/FLMD-DRONIUSF7.config
cvetaevvitaliy: del string set  baro_i2c_address  = ?

PR #239, opened 2020-04-16
Update HGLR-HGLRCF722 config

HGLRC-D:


Asizon:
What about this target issue?
https://github.com/betaflight/betaflight/issues/9685

HGLRC-D:
> 那这个目标问题呢？
> [betaflight / betaflight＃9685](https://github.com/betaflight/betaflight/issues/9685)

We are testing and verifying，we will propose amendments later.

配置/默认/HGLR-HGLRCF722.config
```diff
@@ -50,9 +50,7 @@ resource PINIO 2 C09
 resource FLASH_CS 1 D02
```
mikeller, line 50: You are moving this file into a different directory - this will not work, and the target will become unavailable to users.

PR #245, opened 2020-04-18
Added Unified Target configuration instructions to the pull request template.

mikeller:


etracer65:
Maybe we should add a bullet point about "Before submitting your pull request be sure to thoroughly test the unified target on actual hardware" or something like that. With the frequent significant problems we see it's obvious that some submissions were never tested by the maintainer.

mikeller:
Good point - I will add this to the page with instructions for Unified Target creation. There are a number of places that link to this page already.

PR #253, opened 2020-04-25
Adding RCTimer ARESF7 target configuration.

mikeller:


configs/default/RCTI-ARESF7.config
```diff
@@ -0,0 +1,119 @@
...
+# pin C09: DMA1 Stream 2 Channel 5
+
+# feature
+feature -RX_PARALLEL_PWM
+feature RX_SERIAL
+feature OSD
+
+# serial
+serial 0 64 115200 57600 0 115200
+serial 1 1024 115200 57600 0 115200
```
etracer65, line 94: I'm not sure this makes any sense. Enabling ESC sensor on this UART when the feature isn't even enabled.

configs/default/RCTI-ARESF7.config
```diff
@@ -0,0 +1,119 @@
...
+set osd_craft_name_pos = 2048
+set max7456_spi_bus = 3
+set pinio_config = 129,1,1,1
+set pinio_box = 40,255,255,255
+set flash_spi_bus = 2
+set gyro_1_bustype = SPI
+set gyro_1_spibus = 1
+set gyro_1_sensor_align = CW180
+set gyro_1_align_yaw = 1800
+set name = Ares F7
```
etracer65, line 119: This would be annoying to me (particularly as it's also added to the OSD!), but I guess it's not technically wrong.

PR #255, opened 2020-04-26
Update VGRC-VGOODRCF4.config

hljxkang2020:

### DIFF
--- a/configs/default/VGRC-VGOODRCF4.config
+++ b/configs/default/VGRC-VGOODRCF4.config
@@ -1,24 +1,46 @@
-# Betaflight / STM32F405 (S405) 4.1.0 Apr 30 2019 / 09:30:28 () MSP API: 1.42
+# Betaflight / VGOODRCF4 (VGF4) 4.0.2 May 30 2019 / 10:19:58 () MSP API: 1.41
 
 board_name VGOODRCF4
 manufacturer_id VGRC
 
 # resources
 resource BEEPER 1 C13
-resource MOTOR 1 A09
-resource MOTOR 2 A10
-resource MOTOR 3 B00
-resource MOTOR 4 B01
-resource MOTOR 5 C08
-resource MOTOR 6 C09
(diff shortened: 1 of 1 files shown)


configs/default/VGRC-VGOODRCF4.config
```diff
@@ -1,38 +1,69 @@
...
 
 # resources
 resource BEEPER 1 C13
-resource MOTOR 1 A09
-resource MOTOR 2 A10
-resource MOTOR 3 B00
-resource MOTOR 4 B01
-resource MOTOR 5 C08
-resource MOTOR 6 C09
+resource MOTOR 1 A08
```
mikeller, line 8: Was this board ever sold to customers? If so, this will cause this board to stop working for all existing customers.

configs/default/VGRC-VGOODRCF4.config
```diff
@@ -1,38 +1,69 @@
-# Betaflight / STM32F405 (S405) 4.1.0 Apr 30 2019 / 09:30:28 () MSP API: 1.42
+# Betaflight / VGOODRCF4 (VGF4) 4.0.2 May 30 2019 / 10:19:58 () MSP API: 1.41
```
mikeller, line 1: There are a number of problems with these changes - please read the instructions in https://github.com/betaflight/betaflight/blob/master/docs/TargetMaintenance/CreatingAUnifiedTarget.md and follow them closely to fix the problems.

hljxkang2020: Thanks for your remind. We don't sell  products with old version to the market, as when we test, we find it need to improve, so please replace with this new, thanks.

configs/default/VGRC-VGOODRCF4.config
hljxkang2020: Thanks for your remind. We don't sell  products with old version to the market, as when we test, we find it need to improve, so please replace with this new, thanks.

PR #259, opened 2020-05-02
Added I2C 3 to Matek F405 boards that have it on S4 / S6.

mikeller:
@MATEKSYS are you ok with this change?

### DIFF
--- a/configs/default/MTKS-MATEKF405STD.config
+++ b/configs/default/MTKS-MATEKF405STD.config
@@ -29,7 +29,9 @@ resource SERIAL_RX 3 C11
 resource SERIAL_RX 4 A01
 resource SERIAL_RX 5 D02
 resource I2C_SCL 1 B06
+resource I2C_SCL 3 A08
 resource I2C_SDA 1 B07
+resource I2C_SDA 3 C09
 resource LED 1 B09
 resource LED 2 A14
 resource SPI_SCK 1 A05


MATEKSYS:
I don't feel the need to add I2C3 in these targets.  They all have I2C1 pads.  only F405-OSD board has no I2C1 breakout.   Although there is I2C3 definition in legacy target,  nobody use I2C3 on F405-CTR/STD/mini.

mikeller:
@MATEKSYS:

> Although there is I2C3 definition in legacy target, nobody use I2C3 on F405-CTR/STD/mini.

Apparently some people do: betaflight/betaflight#9750 The user in question (@supersidor) has a MATEKF405-OSD, so I2C 1 is not an option.

MATEKSYS:
I prefer to update MTKS-MATEKF405STD.config only.  so that F405-OSD can use this unified target.

supersidor:
Thanks for efforts, will make firmware updates much easier for me

mikeller:
@MATEKSYS: Updated.

SnowLeopardFPV:
This change has now caused a conflict with motor 4 when using non DSHOT ESC protocols. Please see the comment I added to the bottom of [betaflight/betaflight#9750](https://github.com/betaflight/betaflight/pull/9750).

mikeller:
@SnowLeopardFPV: Can you please open an issue in https://github.com/betaflight/betaflight/issues, and provide the information that you are asked when opening the issue. Also, please refrain from cross-posting the same information across several places - it only makes it less likely that somebody will be wanting to help you.

PR #262, opened 2020-05-07
Update VGRC-VGOODRCF4.config

hljxkang2020:
For the old design hardware, there are some issue, so we stopped and not sell to the market, only some test samples flashed with that version  by ourself. We will use this new updated version only. For this new version, we adjusted the I/O ports and default configs, such like LED light.etc.

### DIFF
--- a/configs/default/VGRC-VGOODRCF4.config
+++ b/configs/default/VGRC-VGOODRCF4.config
@@ -1,18 +1,19 @@
-# Betaflight / STM32F405 (S405) 4.1.0 Apr 30 2019 / 09:30:28 () MSP API: 1.42
+# Betaflight / STM32F405 (S405) 4.1.6 Apr 25 2020 / 05:11:18 (283bda8bf) MSP API: 1.42
 
 board_name VGOODRCF4
 manufacturer_id VGRC
 
 # resources
 resource BEEPER 1 C13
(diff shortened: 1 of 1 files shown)


configs/default/VGRC-VGOODRCF4.config
```diff
@@ -57,70 +57,82 @@ resource USB_DETECT 1 C04
...
-set mag_spi_device = 0
 set baro_bustype = I2C
-set baro_spi_device = 0
 set baro_i2c_device = 2
-set baro_i2c_address = 0
-set adc_device = 2
+set serialrx_provider = SBUS
 set blackbox_device = SPIFLASH
-set dshot_burst = OFF
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 123: If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, potentially resulting in damage or injury.

hljxkang2020: Thanks for your kind remind for the safety.
We sell this controller with our latest BLHeli_32 ESC and our ESC default also with Dshot600, at the same time, we will write and enhance this safety instruction in our FC manual. So we would to keep this default value as this, this is also required by our customer. Thanks for it.

PR #264, opened 2020-05-12
Add GEPRCF7_BT_HD board config

Linjieqiang:


configs/default/GEPR-GEPRCF7_BT_HD.config
```diff
@@ -0,0 +1,153 @@
...
+set gyro_to_use = BOTH
+set baro_bustype = I2C
+set baro_i2c_device = 2
+set mag_bustype = I2C
+set mag_i2c_device = 2
+set serialrx_provider = SBUS
+set adc_device = 3
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 138: If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, potentially resulting in damage or injury.

Linjieqiang: @mikeller Hello, Mikeller, Thank you for your reminder. Our flight controller is equipped with our aircraft, not sold separately. The aircraft uses the standard BLHeli32. So my configuration file should be safe for users.

PR #267, opened 2020-05-14
AIRB-OMNIBUSF4SD - Add missing I2C

etracer65:
Fixes https://github.com/betaflight/betaflight/issues/9802

Board has SCL/SDA markings shared with UART3. In the original legacy target they were noted, but defaulted to NONE.

### DIFF
--- a/configs/default/AIRB-OMNIBUSF4SD.config
+++ b/configs/default/AIRB-OMNIBUSF4SD.config
@@ -50,6 +50,8 @@ resource OSD_CS 1 A15
 resource GYRO_EXTI 1 C04
 resource GYRO_CS 1 A04
 resource USB_DETECT 1 C05
+resource I2C_SCL 2 B10
+resource I2C_SDA 2 B11
 
 # timer
 timer B08 AF3


jflyper:
If resources are defined for I2C, the bus will be configured and activated even without a bus client (a sensor). This is not a desirable effect. Hence the "noted but defaulted to NONE" handling in the legacy target.

etracer65:
@jflyper The goal was to make the bus available to be used. The board has the pads marked "SCL/TX3" and "SDA/RX3". Also the target has:
```
set mag_bustype = I2C
set mag_i2c_device = 2
```
What is the problem having the I2C2 bus active? I tested and UART3 still works if used (tested with smart audio). Also there are 3 other targets defined this way:
```
EXUA-EXF722DUAL
GEPR-GEPRCF7_BT_HD
IFRC-IFLIGHT_F722_TWING
```

mikeller:
@etracer65: I think the problem was / is that, if the pins are not used for an UART, the I2C bus will be initialised, and generate errors that are shown in configurator, resulting in issues being opened by users. Not sure if this is still a problem, and if it is affecting I2C on all pins, as different pins have different internal pull-ups.

etracer65:
Well any time an I2C bus is enabled and MAG or BARO are set to I2C it's going to try and detect devices and generate errors if nothing is connected. This is regardless of whether the I2C pins are shared with a UART or not. So is that really a problem? The presence of a couple of I2C errors at the bottom of the configurator at boot has been there forever. By this logic we shouldn't enable the I2C bus for any pins on any flight controller unless the device is physically on the board and not support external mag/baro.

As mentioned the genuine Airbot board has the pins labeled "SCL/TX2" and "SDC/RX3" so it would seem reasonable that these pins should be expected to work when you connect a mag or baro.

Even though it's been there forever, if the concern is a few I2C errors as unconnected busses are probed then maybe we need to change the firmware to only count the errors after initialization as an indication of an ongoing communication problem.

jflyper:
My concern is backward compatibility. Those who use these pins as I2C do have experience with configuring them explicitly, while those use these pins as UART do not have such experience, and would generate certain amount of confusion.

etracer65:
> My concern is backward compatibility. Those who use these pins as I2C do have experience with configuring them explicitly, while those use these pins as UART do not have such experience, and would generate certain amount of confusion.

But the problem is then we're forcing people to learn how to remap resources to use features that are supposed to work on the board. This PR was raised in response to a support ticket for exactly this reason. The board is labeled that these pins are for connecting an external I2C device so it's reasonable to expect it to work.

Is there actually any concern that the UART won't work correctly if the user connects a serial device like a receiver to it? As already noted, yes there will be a couple of I2C errors as the bus is probed. But that's no different than any other target that has the I2C bus exposed for external devices.

So if there's a technical reason why doing this is bad beyond a couple of inconsequential I2C errors recorded at boot then I'm fine with not doing it. But to not do it because the legacy target didn't support it (incorrectly in my opinion) that doesn't make any sense to me.

jflyper:
> So if there's a technical reason why doing this is bad beyond a couple of inconsequential I2C errors recorded at boot then I'm fine with not doing it. But to not do it because the legacy target didn't support it (incorrectly in my opinion) that doesn't make any sense to me.

Like I said, the legacy target has been regarding I2C on PB10 and PB11 as a user's configuration option for as long as our history of F4 support.  Existing users of this board already have their own fixed usage on how to use these pins, and changing it at this late in the lifecycle of the board would cause some degree of confusion. Assigning them to I2C may benefit new users of this board, but I assume they are minorities compared to the existing user base.

PR #268, opened 2020-05-15
add new target KakuteF7-MINI-V3

jamming:
l was told that BGA package stm32f745 was difficult to buy and expensive. We have to use a 64 pins TQFP  stm32f722 on the newly KakuteF7-Mini series.  The newly version is V3. 

KakuteF7-MINI-V2 which use BGA package stm32f745  was abandoned for mass production.

### DIFF
--- a/configs/default/HBRO-KAKUTEF7MINIV3.config
+++ b/configs/default/HBRO-KAKUTEF7MINIV3.config
@@ -1,0 +1,118 @@
+# Betaflight / STM32F7X2 (S7X2) 4.1.0 Sep  8 2019 / 01:19:23 (f331cb3dd) MSP API: 1.42
+
+board_name KAKUTEF7MINIV3
+manufacturer_id HBRO
+
+# resources
+resource BEEPER 1  C08
+resource MOTOR 1  B00
+resource MOTOR 2  B01
+resource MOTOR 3  B04
+resource MOTOR 4  B05
+resource MOTOR 5  B06
+resource MOTOR 6  B07
+resource PPM 1 A03
+resource LED_STRIP 1 A08
+resource SERIAL_TX 1 A09
+resource SERIAL_TX 2 A02
+resource SERIAL_TX 3 B10
+resource SERIAL_TX 4 A00
+resource SERIAL_TX 6 C06
+resource SERIAL_RX 1 A10
+resource SERIAL_RX 2 A03
+resource SERIAL_RX 3 B11
+resource SERIAL_RX 4 A01
+resource SERIAL_RX 6 C07
+resource I2C_SCL 1 B08
+resource I2C_SDA 1 B09
+resource LED 1 D02
+resource SPI_SCK 1 A05
+resource SPI_SCK 2 B13
+resource SPI_SCK 3 C10
+resource SPI_MISO 1 A06
+resource SPI_MISO 2 B14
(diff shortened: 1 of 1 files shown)


configs/default/HBRO-KAKUTEF7MINIV3.config
```diff
@@ -0,0 +1,118 @@
...
+feature RX_SERIAL
+feature TELEMETRY
+feature OSD
+
+# serial
+serial 1 64 115200 57600 0 115200
+
+# master
+set serialrx_provider = SBUS
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 99: If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, potentially resulting in damage or injury.

configs/default/HBRO-KAKUTEF7MINIV3.config
```diff
@@ -0,0 +1,118 @@
...
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set mag_hardware = NONE
```
mikeller, line 102: This should not be set to `NONE` - the default (`AUTO`) will allow auto-detection of MAG devices if any are connected. With `NONE`, MAG will not be supported on this board.

jamming:
&gt;&gt;+set motor_pwm_protocol = DSHOT600 
&gt;&gt;If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, &gt;&gt;potentially resulting in damage or injury.
&gt;&gt;Do you assume all responsibility for personal injury and damage resulting from using this default setting?

Thank you for your reminder. I will set&nbsp;motor_pwm_protocol to ONESHOT125 as default.



&gt;&gt;+set mag_hardware = NONE 
&gt;&gt;This should not be set to&nbsp;NONE&nbsp;- the default (AUTO) will allow auto-detection of MAG devices if any are connected. With&nbsp;NONE, MAG will not be supported on this board.

There is no onboard MAG on the KakuteF7 series. When setting this parameter to auto, Betaflight Configurator will report several i2c errors. &nbsp;A lot of users are confused about i2c errors.&nbsp;They always send emails asking about it. By&nbsp;&nbsp;setting this parameter to NONE, there will no i2c errors report on Betaflight Configurator. For users who want to connect a external compass,&nbsp;&nbsp;they can manually open it on the configuration page of&nbsp;Betaflight Configurator.







Regards,

--jamming

&nbsp;
&nbsp;
------------------&nbsp;Original&nbsp;------------------

mikeller:
@jamming:

> Thank you for your reminder. I will set&nbsp;motor_pwm_protocol to ONESHOT125 as default.

Best don't set anything at all for this - this will default to 'disabled' after a reset, and configurator will remind users to select the appropriate protocol on connect.
Selecting OneShot125 will result in the PID loop speed being limited to 2 kHz, which also results in a lot of user problem reports.

jamming:
Hi&nbsp;Keller,


&nbsp;It's reasonable not to set this parameter&nbsp;based on this consideration. I will change it and push a new commit.


Thank you!


Regards,
--Jamming
&nbsp;
------------------&nbsp;Original&nbsp;------------------

PR #269, opened 2020-05-15
Add GEPRCF405 unified config file

arbou:
Add GEPRCF405 unified config file for GEPRC

### DIFF
--- a/GEPR-GEPRCF405
+++ b/GEPR-GEPRCF405
@@ -1,0 +1,136 @@
+# Betaflight / STM32F405 (S405) 4.1.6 Apr 25 2020 / 05:11:18 (283bda8bf) MSP API: 1.42
+
+board_name GEPRCF405
+manufacturer_id GEPR
+
+# resources
+resource BEEPER 1 C13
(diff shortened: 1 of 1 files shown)


GEPR-GEPRCF405
```diff
@@ -0,0 +1,136 @@
...
+aux 0 0 0 1700 2100 0 0
+
+# master
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set baro_bustype = I2C
+set baro_i2c_device = 2
+set serialrx_provider = SBUS
+set blackbox_device = SPIFLASH
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 124: If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, potentially resulting in damage or injury.

arbou: Dear Mikeller,
Yes, we will take responsibility, what we do to avoid any injure happen, we will warning customers with this point. Thanks.

PR #273, opened 2020-05-19
Add  GEPR-GEPRCF405 unified targets.

arbou:
Yesterday I have upload files to a wrong directory, I need to recall it and upload files again, sorry for any inconvenience to you. tks.


mikeller:
@arbou: Sorry, my bad for not spotting this. This pull request has got some issues however:
- it creates a new target instead of moving the original one that was added in the wrong location;
- the proposed path is still wrong.

I have created #274 to fix the issue.

PR #275, opened 2020-05-20
add pyrodrone F7 target to manufaturer ID PYRO

bnn1044:
add PYRODRONEF7 TARGET


configs/default/PYRO-PYRODRONEF7.config
```diff
@@ -0,0 +1,115 @@
+# Betaflight / STM32F7X2 (S7X2) 4.1.0 Oct 16 2019 / 00:02:08 (714f27b2d) MSP API: 1.42
```
mikeller, line 1: The file name is wrong - should be `PYDR-PYRODRONEF7.config`.

bnn1044: changed. sorry for the mistake.

PR #287, opened 2020-06-10
add PYRODRONEF4 PDB TARGET

bnn1044:
add PYRODRONEF4 PDB TARGET for menufaturer PYDR

### DIFF
--- a/configs/default/PYDR-PYRODRONEF4PDB.config
+++ b/configs/default/PYDR-PYRODRONEF4PDB.config
@@ -92,7 +92,6 @@ set baro_spi_device = 2
 set blackbox_device = SPIFLASH
 set flash_spi_bus = 2
 set max7456_spi_bus = 2
-set motor_pwm_protocol = DSHOT600
 set current_meter = ADC
 set battery_meter = ADC
 set vbat_scale = 160
--- a/configs/default/PYDR-PYRODRONEF4PDB.config
+++ b/configs/default/PYDR-PYRODRONEF4PDB.config
@@ -45,31 +45,45 @@ resource I2C_SCL 2 B10
 resource I2C_SDA 2 B11
 
 # timer
-timer B09 AF3           # pin B09: TIM11 CH1 (AF3)
-timer B01 AF2           # pin B01: TIM3 CH4 (AF2)
-timer B00 AF3           # pin B00: TIM8 CH2N (AF3)
-timer C09 AF3           # pin C09: TIM8 CH4 (AF3)
-timer A08 AF1           # pin A08: TIM1 CH1 (AF1)
-timer B08 AF2           # pin B08: TIM4 CH3 (AF2)
+# pin B09: TIM11 CH1 (AF3)
+timer B09 AF3           
+# pin B01: TIM3 CH4 (AF2)
+timer B01 AF2           
+# pin B00: TIM8 CH2N (AF3)
+timer B00 AF3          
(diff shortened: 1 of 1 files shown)


configs/default/PYDR-PYRODRONEF4PDB.config
```diff
@@ -0,0 +1,90 @@
...
+dma pin C09 0           # pin C09: DMA2 Stream 7 Channel 7
+dma pin A08 1           # pin A08: DMA2 Stream 1 Channel 6
+dma pin B08 0           # pin B08: DMA1 Stream 7 Channel 2
+
+# feature
+feature OSD
+
+# aux
+aux 0 40 255 900 2100 0 0               // turn on vtx PIT switch 
+aux 1 41 255 900 2100 0 0               // turn on BT module
```
etracer65, line 66: Comments are not supported on the ends of commands. It may accidentally work in some cases but may break in the future. If you want to add comments then do so on lines starting with `#`.

configs/default/PYDR-PYRODRONEF4PDB.config
```diff
@@ -0,0 +1,90 @@
...
+
+
+# serial
+serial 3 1 19200 57600 0 115200        // set up for BT module
+
+
+# master
+set mag_hardware = NONE
```
etracer65, line 74: There is no reason to set this to `NONE`. It will just make it harder for anybody who might want to add an external I2C magnetometer in the future.

Speaking of which, I notice you haven't defined any I2C bus pins. Since I2C2 is an alternate function for USART3 you could add:
```
resource I2C_SCL 2 B10
resource I2C_SDA 2 B11
```
to make it easier for customers. It's a common support question about how to add a magnetometer via I2C and many board targets make it hard for the customer. Additionally you can add:
```
set mag_i2c_device = 2
```
Then all the user would need to do is connect the SCL/SDA pins to UART3 and the magnetometer will work.

bnn1044: got it. making change now. thanks for the suggestion.

configs/default/PYDR-PYRODRONEF4PDB.config
mikeller: @bnn1044: You still need to remove the comments at the end of the lines.

bnn1044: > @bnn1044: You still need to remove the comments at the end of the lines.

changed. thanks

configs/default/PYDR-PYRODRONEF4PDB.config
```diff
@@ -0,0 +1,106 @@
...
+serial 3 1 19200 57600 0 115200       
+
+
+# master
+set mag_i2c_device = 2
+set baro_spi_device = 2
+set blackbox_device = SPIFLASH
+set flash_spi_bus = 2
+set max7456_spi_bus = 2
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 95: If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, potentially resulting in damage or injury.

bnn1044: Ok. Sorry my mistake. I will take that out. Since this is with PDB. People may have some old esc laying around to use. I change that.

PR #288, opened 2020-06-12
add HOBBYWING_XROTORF4G3

Project399:
add unified target config for Hobbywing xRotor F4 G3 flight controller

First time doing pull request, I have followed all guidelines, but please let me know if there is anything I need to amend


configs/default/HOWI-HOBBYWING_XROTORF4G3.config
```diff
@@ -0,0 +1,144 @@
+# Betaflight / STM32F405 (S405) 4.1.7 May 28 2020 / 15:05:17 (9ba02a587) MSP API: 1.42
+# manufacturer_id: HOWI   board_name: HOBBYWING_XROTORF4G3   custom defaults: NO
```
mikeller, line 2: Please remove the line starting with `# manufacturer_id`.

etracer65:
Seems like the board name should be `XROTORF4G3` and not include the manufacturer name to be consistent with every other board.

Project399:
Understood mikeller, I've removed that line. 

As for the naming, I followed what many popular boards use for their naming scheme. For example, some names that are currently in the unified-config list: AIKONF4, BETAFPVF405, IFLIGHT_F411_PRO, MAYTEMF405CTR, SPRACINGF4NEO and so on.  These all have the company name attached to the board name. I believe it would make it much much easier for the customers to be able to search for Hobbywing, than xRotor.

PR #291, opened 2020-06-15
Add VGRC-VGOODF722DUAL unified target

hljxkang2020:
Add VGRC-VGOODF722DUAL unified target

### DIFF
--- a/configs/default/VGRC-VGOODF722DUAL.config
+++ b/configs/default/VGRC-VGOODF722DUAL.config
@@ -1,0 +1,143 @@
+# Betaflight / STM32F7X2 (S7X2) 4.1.7 May 28 2020 / 15:06:20 (9ba02a587) MSP API: 1.42
+# manufacturer_id: VGRC   board_name: VGOODF722DUAL   custom defaults: NO
+
+board_name VGOODF722DUAL
+manufacturer_id VGRC
+
+# resources
+resource BEEPER 1 C13
+resource MOTOR 1 B04
+resource MOTOR 2 B05
+resource MOTOR 3 B00
+resource MOTOR 4 B01
+resource MOTOR 5 B06
+resource MOTOR 6 B07
+resource PPM 1 A03
+resource PWM 1 A02
+resource LED_STRIP 1 A08
+resource SERIAL_TX 1 A09
+resource SERIAL_TX 2 A02
+resource SERIAL_TX 4 A00
+resource SERIAL_TX 6 C06
+resource SERIAL_RX 1 A10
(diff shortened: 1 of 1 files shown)


configs/default/VGRC-VGOODF722DUAL.config
```diff
@@ -0,0 +1,143 @@
+# Betaflight / STM32F7X2 (S7X2) 4.1.7 May 28 2020 / 15:06:20 (9ba02a587) MSP API: 1.42
+# manufacturer_id: VGRC   board_name: VGOODF722DUAL   custom defaults: NO
```
mikeller, line 2: Remove this line.

configs/default/VGRC-VGOODF722DUAL.config
```diff
@@ -0,0 +1,143 @@
...
+led 4 4,15::ATI:0
+led 5 5,15::ATI:0
+led 6 6,15::ATI:0
+
+# aux
+aux 0 0 0 1700 2100 0 0
+aux 1 41 3 1400 2100 0 0
+
+# master
+set acc_calibration = -32,115,-526
```
mikeller, line 119: This will result in a faulty acc calibration for most users.

configs/default/VGRC-VGOODF722DUAL.config
```diff
@@ -0,0 +1,143 @@
...
+set mag_bustype = I2C
+set mag_i2c_device = 2
+set baro_bustype = I2C
+set baro_i2c_device = 1
+set serialrx_provider = SBUS
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 127: If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, potentially resulting in damage or injury.

configs/default/VGRC-VGOODF722DUAL.config
hljxkang2020:
```suggestion

```

hljxkang2020: Hi Mikeller, 
Thanks for reminding me!  I have removed it.

configs/default/VGRC-VGOODF722DUAL.config
hljxkang2020: Hi Mikeller,
Thanks for reminding this, we noted and will keep this as it is, we will warn this in our manual and will take responsibility for it.

PR #292, opened 2020-06-16
Update HBRO-KAKUTEF7MINI.config

theely:
Fix for #9916 enabling camera control
Please note some comments need fixing, I don't know where to retrieve information about DMA and TIM1 channels


mikeller:
@theely: You can get these from a `dump hardware`:

```
# dump hardware

[...]
resource CAMERA_CONTROL 1 B03
[...]
timer B03 AF1
# pin B03: TIM2 CH2 (AF1)
[...]
dma pin B03 0
# pin B03: DMA1 Stream 6 Channel 3
[...]
```

Also, camera control does not use DMA, so not sure why the manufacturer says that this is needed, `dma pin B03 0` can probably be removed.

PR #295, opened 2020-06-22
Add GEPRCF722 Unified Targets for GEPRC.

arbou:
Add GEPRCF722 Unified Targets for GEPRC.
For the  safety about default with Dshot 600:
We sell this controller with our latest BLHeli_32 ESC and our ESC default also with Dshot600, at the same time, we will write and enhance this safety instruction in our FC manual. We will take responsibility for any safety issue if there. tks.


mikeller:
This seems to be missing the files. :-)

PR #302, opened 2020-07-01
add new unified target for SJMT_AOCODAF722BLE

star624:
The new unified target for our F7 flight controller, thanks for your work!


Asizon:
@star624, Wrong file name, you need to use "-" and not "_" after manufacturer ID

star624:
i'm very sorry for my carelessness,i have modified the error.

PR #304, opened 2020-07-03
add new unified target for SJET_AOCODAF4

star624:
add a new unified target for our F4 flight controller, thanks for your work!


star624:
yes ,i have rename file and board_name, thank you!

star624:
rename the file and the board_name in file.

PR #310, opened 2020-07-08
Added target EACC-EACHINEF722

Eachine-D:


configs/default/EACC-EACHINEF722.config
```diff
@@ -0,0 +1,132 @@
+# Betaflight / STM32F405 (S405) 4.2.0 Dec  4 2019 / 01:50:58 (ebdda3dec7) MSP API: 1.43
+
+board_name EACHINEF722
+manufacturer_id EACC
```
Asizon, line 4: Non existent manufacturer_id, please follow unified targets instructions to REQUEST it.

PR #315, opened 2020-07-11
Change LDARCF411 fc board config.

Linjieqiang:
According to the manufacturer's requirements.

### DIFF
--- a/configs/default/LDAR-LDARCF411.config
+++ b/configs/default/LDAR-LDARCF411.config
@@ -4,10 +4,12 @@ manufacturer_id LDAR
 
 # resources
 resource BEEPER 1 B02
-resource MOTOR 1 A00
-resource MOTOR 2 B10
+resource MOTOR 1 B04
+resource MOTOR 2 B05
 resource MOTOR 3 B06
 resource MOTOR 4 B07
+resource MOTOR 5 B03
+resource MOTOR 6 B10
 resource PPM 1 A03
 resource LED_STRIP 1 A08
 resource SERIAL_TX 1 A09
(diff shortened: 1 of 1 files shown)


configs/default/LDAR-LDARCF411.config
```diff
@@ -20,63 +22,96 @@ resource LED 1 C13
...
-resource SPI_MISO 3 B04
 resource SPI_MOSI 1 A07
 resource SPI_MOSI 2 B15
-resource SPI_MOSI 3 B05
 resource ADC_BATT 1 B00
 resource ADC_CURR 1 B01
-resource FLASH_CS 1 A15
 resource OSD_CS 1 B12
 resource GYRO_EXTI 1 A01
+resource GYRO_EXTI 2 NONE
```
mikeller, line 33: This is not needed.

configs/default/LDAR-LDARCF411.config
```diff
@@ -4,10 +4,12 @@ manufacturer_id LDAR
 
 # resources
 resource BEEPER 1 B02
-resource MOTOR 1 A00
-resource MOTOR 2 B10
+resource MOTOR 1 B04
```
mikeller, line 7: Has this board ever been sold in the existing configuration? If so, this should be added as a new target, and marketed as a separate version of the board.

configs/default/LDAR-LDARCF411.config
```diff
@@ -20,63 +22,96 @@ resource LED 1 C13
...
+serial 0 64 115200 57600 0 115200
 
 # master
+set serialrx_provider = SBUS
+set dshot_burst = AUTO
+set dshot_bitbang = OFF
+set motor_pwm_protocol = DSHOT600
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set mag_hardware = NONE
```
mikeller, line 101: These settings don't make sense - either do not configure I2C for mag, or set `mag_hardware` to `auto` to enable auto-detection of devices.

configs/default/LDAR-LDARCF411.config
```diff
@@ -20,63 +22,96 @@ resource LED 1 C13
...
 set baro_bustype = I2C
 set baro_i2c_device = 1
-set blackbox_device = SPIFLASH
+set baro_hardware = NONE
```
mikeller, line 104: If this is set, users will be told that baro is not supported on this board, and to buy a new board if they want to use baro.

configs/default/LDAR-LDARCF411.config
```diff
@@ -20,63 +22,96 @@ resource LED 1 C13
...
+feature DYNAMIC_FILTER
+
+# serial
+serial 0 64 115200 57600 0 115200
 
 # master
+set serialrx_provider = SBUS
+set dshot_burst = AUTO
+set dshot_bitbang = OFF
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 98: If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, potentially resulting in damage or injury.  

configs/default/LDAR-LDARCF411.config
Linjieqiang: Hello mikeller. I also received a request from this manufacturer to make the above modifications. Their products should not have been sold.

configs/default/LDAR-LDARCF411.config
Linjieqiang: In this situation, the manufacturer has understood and made it clear that it is willing to pay full responsibility.

PR #316, opened 2020-07-13
Added target EACH-EACHINEF722

EachineRC:


configs/default/EACH-EACHINEF722.config
```diff
@@ -0,0 +1,132 @@
+# Betaflight / STM32F405 (S405) 4.2.0 Dec  4 2019 / 01:50:58 (ebdda3dec7) MSP API: 1.43
```
mikeller, line 1: Please test your targets with released firmware and not with development builds.

configs/default/EACH-EACHINEF722.config
```diff
@@ -0,0 +1,132 @@
...
+# serial
+serial 0 64 115200 57600 0 115200
+
+# master
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set baro_spi_device = 1
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 119: If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, potentially resulting in damage or injury.  

EachineRC: Thank you very much for your reminder. We have gone through internal consensus and decided not to make changes. We will make instructions for the use of products for our customers.

EachineRC:
> Before we start adding new targets for you, can I please ask you to get compliant with the Betaflight licensing conditions, and publish the source for the firmware you are shipping with the following targets, either by opening pull requests for legacy targets in Betaflight, or by publishing the sources on the support pages for these models:
> 
> * Eachine E010S (see [betaflight/betaflight#6828](https://github.com/betaflight/betaflight/issues/6828));
> * Eachine Chaser88 (see [betaflight/betaflight#4544](https://github.com/betaflight/betaflight/issues/4544)).

Sorry, we don’t know about the pull request for the old version of the target, we have not made the relevant request before, and they cannot represent EACHINE.

mikeller:
@EachineRC: There are no pull requests for open source firmware for these products, which is why they are breaking the licensing conditions of Betaflight:

https://www.eachine.com/Eachine-E010S-65mm-Micro-FPV-Racing-Quadcopter-with-800TVL-CMOS-Based-On-F3-Brush-Flight-Controller-p-600.html

https://transcendfpv.com/eachine-chaser88-review/

This licensing breach needs to be rectified by the publication of the source code for the Betaflight firmware supplied with these boards before more targets from this manufacturer can be accepted.

mikeller:
@EachineRC: Thanks for providing the information to make the Eachine E010S compliant with Betaflight's license in https://github.com/betaflight/betaflight/issues/6828#issuecomment-659222287. We'll have to work with owners of Eachine Chaser88 to see what can be done there.

configs/default/EACH-EACHINEF722.config
mikeller: @EachineRC: Does this mean you assume all responsibility for personal injury and damage resulting from using this default setting?

mikeller: Can I get you to confirm in the comment on the Dshot600 default that you assume all responsibility for damage and injury resulting from this?

1:23 PM￼
Doven Li:
Yes

PR #318, opened 2020-07-16
Add TRANSTECF405HD

flosean:
Add a new target to support TransTEC new F405 + Blheli_S AIO.
This is only for the DJI FPV system.

### DIFF
--- a/configs/default/TTRH-TRANSTECF405HD.config
+++ b/configs/default/TTRH-TRANSTECF405HD.config
@@ -93,7 +93,6 @@ set baro_hardware = NONE
 set serialrx_provider = SBUS
 set blackbox_device = NONE
 set dshot_burst = ON
-set motor_poles = 12
 set motor_pwm_protocol = DSHOT600
 set current_meter = ADC
 set battery_meter = ADC
--- a/configs/default/TTRH-TRANSTECF405HD.config
+++ b/configs/default/TTRH-TRANSTECF405HD.config
@@ -1,0 +1,109 @@
+# Betaflight / STM32F405 (S405) 4.2.0 Jun 14 2020 / 03:04:22 (8f2d21460) MSP API: 1.43
+board_name TRANSTECF405HD
+manufacturer_id TTRH
+
+# name: TransTEC
+
+# resources
+resource BEEPER 1 B04
+resource MOTOR 1 C06
+resource MOTOR 2 C07
+resource MOTOR 3 C08
+resource MOTOR 4 C09
+resource LED_STRIP 1 B06
+resource SERIAL_TX 1 A09
+resource SERIAL_TX 2 A02
+resource SERIAL_TX 3 B10
+resource SERIAL_TX 4 A00
+resource SERIAL_TX 5 C12
+resource SERIAL_RX 1 A10
+resource SERIAL_RX 2 A03
+resource SERIAL_RX 3 B11
+resource SERIAL_RX 4 A01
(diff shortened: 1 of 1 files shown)


configs/default/TTRH-TRANSTECF405HD.config
```diff
@@ -0,0 +1,109 @@
...
+aux 1 13 1 1700 2100 0 0
+
+# master
+set mag_hardware = NONE
+set baro_hardware = NONE
+set serialrx_provider = SBUS
+set blackbox_device = NONE
+set dshot_burst = ON
+set motor_poles = 12
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 97: If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, potentially resulting in damage or injury.

configs/default/TTRH-TRANSTECF405HD.config
```diff
@@ -0,0 +1,109 @@
...
+led 7 7,0::C:1
+led 8 8,0::C:1
+
+# aux
+aux 0 0 2 1300 2100 0 0
+aux 1 13 1 1700 2100 0 0
+
+# master
+set mag_hardware = NONE
+set baro_hardware = NONE
```
mikeller, line 92: If this is set, users will be told that baro is not supported on this board, and to buy a new board if they want to use baro.

configs/default/TTRH-TRANSTECF405HD.config
```diff
@@ -0,0 +1,109 @@
...
+led 6 6,0::C:1
+led 7 7,0::C:1
+led 8 8,0::C:1
+
+# aux
+aux 0 0 2 1300 2100 0 0
+aux 1 13 1 1700 2100 0 0
+
+# master
+set mag_hardware = NONE
```
mikeller, line 91: If this is set, users will be told that mag is not supported on this board, and to buy a new board if they want to use mag.

configs/default/TTRH-TRANSTECF405HD.config
```diff
@@ -0,0 +1,109 @@
...
+set baro_hardware = NONE
+set serialrx_provider = SBUS
+set blackbox_device = NONE
+set dshot_burst = ON
+set motor_poles = 12
```
mikeller, line 96: This is an unusual default. Will this be sold only in an RTF offer with motors?

configs/default/TTRH-TRANSTECF405HD.config
flosean: Hi, Mikeller
This board is a AIO board, FC will only come with the ESC. And most of the drone will come with RTF.

configs/default/TTRH-TRANSTECF405HD.config
flosean: Thanks for remind, have deleted the default motor poles setting.
Forget there will have different setting in 110X and 140X motor.

configs/default/TTRH-TRANSTECF405HD.config
flosean: This board will not support baro for now😢

configs/default/TTRH-TRANSTECF405HD.config
flosean: This board will not support mag for now😢

PR #320, opened 2020-07-24
Added_FLCO_FLYCOLORF7_config

tobeornottobe81:


configs/default/FLCO-FLYCOLORF7.config
```diff
@@ -0,0 +1,144 @@
...
+
+# master
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set baro_bustype = I2C
+set baro_i2c_device = 1
+set adc_device = 3
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = ONESHOT125
```
mikeller, line 126: This seems to be an odd choice, as it will limit the default PID loop speed that will be set for users, forcing them to change this, and then change the PID loop speed.

tobeornottobe81: Hi,mikeller.
For some reasons, such as beginners or ESC does not support it, dshot is not selected as the default protocol, and customers can choose it by themselves,thank you.

mikeller: This is a poor solution, and will result in unnecessary support cases being opened about the PID loop speed being lower than other boards.
As stated above the correct approach is to remove this line, in which case the ESC protocol will be set to 'none', and the user asked to select the appropriate protocol when they connect for the first time.
If you decide to go with the current configuration, in order to minimise the effort our volunteers have to put into answering unnecessary support calls for your hardware, the standard answer for any questions pertaining to PID loop speeds will be that more than 2 kHz is not supported by Betaflight for this board, and that users will have to buy different hardware if they want to change this.

tobeornottobe81: I'll correct it right away. Please wait a moment

configs/default/FLCO-FLYCOLORF7.config
```diff
@@ -0,0 +1,144 @@
...
+
+# master
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set baro_bustype = I2C
+set baro_i2c_device = 1
+set adc_device = 3
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 126: If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, potentially resulting in damage or injury.

tobeornottobe81: hi mikeller，Thank you for your attention.
In theory, what you say will happen. Where there are dangers, we will emphasize the use methods and precautions in the manual, but playing with the aircraft itself is dangerous and needs special attention. in addition
1. Our flight control firmware has been tested for 4-5 months, The effect is very satisfactory
2. The current ESC basically supports dshot
3. We are responsible for all personal injury caused by using this setting
thank you!

PR #327, opened 2020-08-02
Add IFRC-JBF7 serial flight controller support.

Linjieqiang:
The Original from [JBF7 confirmed targets.zip](https://github.com/betaflight/unified-targets/files/5012101/JBF7.confirmed.targets.zip)


mikeller:
Please read https://github.com/betaflight/betaflight/blob/master/docs/TargetMaintenance/CreatingAUnifiedTarget.md for instructions on how to add new Unified Target configurations.

Linjieqiang:
Submit as required.

configs/default/IFRC-JBF7.config
```diff
@@ -0,0 +1,213 @@
...
+# vtxtable
+vtxtable bands 1
+vtxtable channels 8
+vtxtable band 1 RACEBAND R FACTORY    0    0 5732 5769 5806 5843    0    0
+vtxtable powerlevels 1
+vtxtable powerlabels 25
+
+# master
+set gyro_to_use = BOTH
+set acc_calibration = -32,-99,-12,1
```
mikeller, line 128: This should not be in here - this will prevent users from properly calibrating their ACC, potentially resulting in damage or injury from a runaway when activating any of the stabilised modes with a miscalibrated ACC.

configs/default/IFRC-JBF7.config
```diff
@@ -0,0 +1,213 @@
...
+set mag_hardware = NONE
+set baro_bustype = I2C
+set baro_i2c_device = 2
+set baro_hardware = NONE
+set serialrx_provider = SBUS
+set adc_device = 3
+set blackbox_p_ratio = 64
+set blackbox_device = SDCARD
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 140: If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, potentially resulting in damage or injury.

configs/default/IFRC-JBF7_DJI.config
```diff
@@ -0,0 +1,176 @@
...
+set baro_i2c_device = 2
+set baro_hardware = NONE
+set min_check = 1010
+set max_check = 1990
+set serialrx_provider = SBUS
+set adc_device = 3
+set blackbox_p_ratio = 64
+set blackbox_device = SDCARD
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 138: If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, potentially resulting in damage or injury.

configs/default/IFRC-JBF7_DJI.config
```diff
@@ -0,0 +1,176 @@
...
+beacon RX_SET
+
+# serial
+serial 0 64 115200 57600 0 115200
+serial 1 1 115200 57600 0 115200
+serial 3 1024 115200 57600 0 115200
+
+# master
+set gyro_to_use = BOTH
+set acc_calibration = -32,-99,-12,1
```
mikeller, line 124: This should not be in here - this will prevent users from properly calibrating their ACC, potentially resulting in damage or injury from a runaway when activating any of the stabilised modes with a miscalibrated ACC.

configs/default/IFRC-JBF7.config
```diff
@@ -0,0 +1,213 @@
...
+vtxtable band 1 RACEBAND R FACTORY    0    0 5732 5769 5806 5843    0    0
+vtxtable powerlevels 1
+vtxtable powerlabels 25
+
+# master
+set gyro_to_use = BOTH
+set acc_calibration = -32,-99,-12,1
+set mag_bustype = I2C
+set mag_i2c_device = 2
+set mag_hardware = NONE
```
mikeller, line 131: If this setting is used, this means that users will not be able to use any magnetometer on this board, and if they open support requests about this they will be told they will have to buy a different flight controller if they want to use a magnetometer.

configs/default/IFRC-JBF7_DJI.config
```diff
@@ -0,0 +1,176 @@
...
+set mag_bustype = I2C
+set mag_i2c_device = 2
+set mag_hardware = NONE
```
mikeller, line 127: If this setting is used, this means that users will not be able to use any magnetometer on this board, and if they open support requests about this they will be told they will have to buy a different flight controller if they want to use a magnetometer.

configs/default/IFRC-JBF7_DJI.config
```diff
@@ -0,0 +1,176 @@
...
+set baro_bustype = I2C
+set baro_i2c_device = 2
+set baro_hardware = NONE
```
mikeller, line 130: If this setting is used, this means that users will not be able to use any barometer on this board, and if they open support requests about this they will be told they will have to buy a different flight controller if they want to use a barometer.

configs/default/IFRC-JBF7.config
```diff
@@ -0,0 +1,213 @@
...
+set baro_bustype = I2C
+set baro_i2c_device = 2
+set baro_hardware = NONE
```
mikeller, line 134: If this setting is used, this means that users will not be able to use any barometer on this board, and if they open support requests about this they will be told they will have to buy a different flight controller if they want to use a barometer.

configs/default/IFRC-JBF7.config
```diff
@@ -0,0 +1,213 @@
...
+set serialrx_provider = SBUS
+set adc_device = 3
+set blackbox_p_ratio = 64
+set blackbox_device = SDCARD
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
+set vbat_max_cell_voltage = 440
+set current_meter = ESC
```
mikeller, line 142: Is this only to be sold as an all-in-one with ESCs that support current measuring?
Users will not be able to use the on-board current sensor, and the resources for it should be removed from the target, as this is misleading.

configs/default/IFRC-JBF7_DJI.config
```diff
@@ -0,0 +1,176 @@
...
+set min_check = 1010
+set max_check = 1990
+set serialrx_provider = SBUS
+set adc_device = 3
+set blackbox_p_ratio = 64
+set blackbox_device = SDCARD
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
+set vbat_max_cell_voltage = 440
+set current_meter = ESC
```
mikeller, line 140: Is this only to be sold as an all-in-one with ESCs that support current measuring?
Users will not be able to use the on-board current sensor, and the resources for it should be removed from the target, as this is misleading.

configs/default/IFRC-JBF7.config
Linjieqiang: Removed it, thanks for reminding.

configs/default/IFRC-JBF7_DJI.config
Linjieqiang: Removed it, thanks for reminding.

configs/default/IFRC-JBF7_DJI.config
Linjieqiang: We have indicated the risk in the product instructions, and we are willing to bear all risks related to it.

configs/default/IFRC-JBF7.config
Linjieqiang: We have indicated the risk in the product instructions, and we are willing to bear all risks related to it.

configs/default/IFRC-JBF7.config
Linjieqiang: We have indicated the risk in the product instructions, and we are willing to bear all risks related to it.

configs/default/IFRC-JBF7.config
Linjieqiang: Yes, this product does not have a magnetic compass sensor.

configs/default/IFRC-JBF7_DJI.config
Linjieqiang: Yes, this product does not have a magnetic compass sensor.

configs/default/IFRC-JBF7.config
Linjieqiang: Yes, this product does not have a baro sensor.

configs/default/IFRC-JBF7_DJI.config
Linjieqiang: Yes, this product does not have a baro sensor.

configs/default/IFRC-JBF7_DJI.config
Linjieqiang: Yes, this fc board is be sold as an all-in-one with ESCs that support current measuring.

joshuabardwell: The board will be sold with ESC that supports current sensing, hence we would like the FC to default to ESC current meter. However the user should be able to change to On-Board ADC if they are using an ESC with analog current sense. Will the submitted config accomplish this?

configs/default/IFRC-JBF7.config
joshuabardwell: It would be acceptable to change baro_hardware to AUTO in case there was an edge case where a user wanted to use some external baro.

configs/default/IFRC-JBF7_DJI.config
joshuabardwell: It would be acceptable to set mag_hardware to AUTO in case the user decided to use a compass via i2c interface.

configs/default/IFRC-JBF7.config
```diff
@@ -0,0 +1,212 @@
...
+set gyro_2_align_yaw = 900
+
+# rateprofile 0
+set roll_rc_rate = 127
+set pitch_rc_rate = 127
+set roll_expo = 40
+set pitch_expo = 40
+set roll_srate = 72
+set pitch_srate = 72
+set yaw_srate = 75
```
etracer65, line 212: Including rates in a target config is a bad idea. Besides being a personal preference setting, many people fly with default rates and this flight controller would behave differently since it has different default rates.

configs/default/IFRC-JBF7.config
```diff
@@ -0,0 +1,212 @@
...
+serial 0 64 115200 57600 0 115200
+serial 3 1024 115200 57600 0 115200
+serial 4 2048 115200 57600 0 115200
+
+# vtxtable
+vtxtable bands 1
+vtxtable channels 8
+vtxtable band 1 RACEBAND R FACTORY    0    0 5732 5769 5806 5843    0    0
+vtxtable powerlevels 1
+vtxtable powerlabels 25
```
etracer65, line 124: Including the vtxtable in the unified config is a bad idea. This presumes that the user will only have a smartaudio VTX as the table is not compatible with Tramp. Further are iFlight and/or JB going to indemnify Betaflight against legal action and guarantee that this table is legal worldwide?

configs/default/IFRC-JBF7_DJI.config
```diff
@@ -0,0 +1,175 @@
...
+set gyro_2_sensor_align = CW90
+set gyro_2_align_yaw = 900
+
+set roll_rc_rate = 127
+set pitch_rc_rate = 127
+set roll_expo = 40
+set pitch_expo = 40
+set roll_srate = 72
+set pitch_srate = 72
+set yaw_srate = 75
```
etracer65, line 175: Same comment that the rates are not appropriate to be in the unified target config.

configs/default/IFRC-JBF7.config
mikeller: Agreed on this point - this will not work for any Tramp user, as tramp does not support 'FACTORY' bands at all - this will give 27% of VTX users a non-working configuration:
image

joshuabardwell: We very much want to ship with a working vtxtable for the largest number of users. If the user must change the vtxtable because they have some other vTX, then we can walk them through the process.

If we change from "factory" to "non-factory" then Tramp still won't work since power value is not correct--right? But shipping with SA2.0 compatible vtxtable I think will provide working config for the largest number of users by default.

iFlight should speak to indemnification and risk however I believe we have researched and consider these channels to be acceptable to ship. We understand that BF project prefers to ship with un-configured vtxtable because of potential legal risk, and that we are explicitly changing that decision of our own volition.

etracer65: The problem is that this `vtxtable` won't work for SA either. There's a misunderstanding about how the bands work I think. SA can change channels in two different ways. The first is by band/channel (FACTORY mode). In this mode the vtx is sent a band and channel number like 1 and 4. **The actual frequencies in the `vtxtable` are meaningless and are simply labels for the user.** The vtx decides what actual frequency to set based on its internal mapping of band/channel to frequency. Every SA vtx supports this.

The next mode is "CUSTOM". When the band is defined this way then the actual frequency from the `vtxtable` is sent and the band/channel is irrelevant. So in this case the frequency in the table does matter and is the controlling factor. The problem is that this mode is not supported by many SA "compatible" vtx's. And if it is supported then it's only partially supported in that sending a frequency not in the vtx's internal table will generally be ignored (some shut off, others will go to the lowest frequency, or some other unexpected behavior). Real TBS vtx's support the frequency mode correctly and are the only ones that can be counted on for this behavior.

So the problem with the approach in this target is that to most vtx's band "1" is not race band. It's often BOSCAM A or sometimes FATSHARK band. It just depends on how the vtx manufacturer implemented their internal table. So you might think, "well we'll just make it band 5 - that's race band right?". Well no, not necessarily as the 5th band is not guaranteed to be race band. And to do that you'd need to define the vtxtable with 5 bands and it'd look like this:
```
vtxtable bands 5
vtxtable channels 8
vtxtable band 1 BAND1    1 CUSTOM     0    0    0    0    0    0    0    0
vtxtable band 2 BAND2    2 CUSTOM     0    0    0    0    0    0    0    0
vtxtable band 3 BAND3    3 CUSTOM     0    0    0    0    0    0    0    0
vtxtable band 4 BAND4    4 CUSTOM     0    0    0    0    0    0    0    0
vtxtable band 5 RACEBAND R FACTORY    0    0 5732 5769 5806 5843    0    0
vtxtable powerlevels 1
vtxtable powervalues 0
vtxtable powerlabels 25
```
And this would obviously be very confusing (and still not guaranteed to work). And as stated above you can't choose to make the band "CUSTOM" as that's not supported by many non-TBS SA vtx's.

Just a side note that Tramp protocol **only** supports CUSTOM and setting by frequency and doesn't have a band/channel capability. But even then there are a number of different implementations and the vtx won't always set any arbitrary frequency. Some will only set to actual channels as defined in the vtx internally and will stop transmitting if some other value is set. Matek vtx's work this way for example.

joshuabardwell: Do you have a recommendation for how to provide a working vtxtable for the largest number of users? You can probably infer from the vtxtable submitted the channel set that we consider sufficiently legal for worldwide use. I understand that Betaflight project and devs are not willing to assume liability for this decision. But there's nobody more qualified to provide technical advice.

joshuabardwell: What about:

vtxtable band 1 RACEBAND R CUSTOM 0 0 5732 5769 5806 5843 0 0

mikeller: As @etracer65 explains above, many non-TBS SmartAudio VTX do not support `CUSTOM` - so even this would leave a considerable part of users with a configuration that looks ok but does not work with their hardware.
The problem of providing a working VTXTABLE for any given VTX hardware is really a task that is up to the manufacturer of the hardware. Think of this just like you think of the drivers that are needed to make peripherals that you plug into you PC work - they are bits of code that are specific to the hardware you are connecting, and only the manufacturer will have all of the information required to fully support the hardware. If you buy a new Logitech webcam and plug it into your Lenovo PC, and it does not work, you'll expect Logitech to supply you with a driver, and you won't go and write to Lenovo and demand that they give you a driver for your Logitech device, or will you?

etracer65: @joshuabardwell The unified target config is meant to describe and configure the **hardware** of the flight controller to make it functional. It is **not** meant to set user preferences, branding, or other configuration shortcuts. This is why I'm opposed to the rates preferences being included - they have nothing to do with hardware configuration of the flight controller.

Likewise beyond the problems of it being impossible to supply a "universal" vtxtable as discussed above, it's again against the point of the unified target config. It's not meant to configure peripherals that the user may or may not connect. Additionally the vtx channels, frequencies, power levels, etc. were intentionally removed from Betaflight and replaced with vtxtables requiring user action to provide protection from sanction and possible legal actions from governmental agencies as you're aware. Just looking at recent events with massive fines of Hobbyking and earlier GetFPV reinforce this position. So trying to add this back via a unified target config is really not acceptable and in my opinion and an attempt to backdoor our policies.

joshuabardwell: Understood. I have requested iFlight to remove the vtxtable lines and to remove the custom rates lines from the supplied config.

Regarding "configure hardware of FC" vs. "configure peripherals" one point of confusion is that FC targets typically lay out the Ports tab for the expected use of UARTS, such as serial RX enabled on one UART, SmartPort enabled on another, etc. This was the main reason for submitting two separate targets--one for DJI and one for analog.

There hasn't been any objection so far to delivering the target with a custom OSD configuration (one for DJI one for analog). Is that acceptable or does that also fall under user preference and need to be removed?

etracer65: > Regarding "configure hardware of FC" vs. "configure peripherals" one point of confusion is that FC targets typically lay out the Ports tab for the expected use of UARTS, such as serial RX enabled on one UART, SmartPort enabled on another, etc. This was the main reason for submitting two separate targets--one for DJI and one for analog. To save the user the step of configuring their own Ports tab.

I don't have a problem with this. If the configuration matches the labeling on the board then in my opinion it's part of configuring the hardware. Like if the SBUS pad is connected to UART1 then selecting serial rx for that is logical. Or a pad for smartport telemetry and having that UART set with telemetry enabled. Of course the downside to this is if the user want's to do a different setup it often leads to confusion in them thinking they can't use those pads for some other UART-based feature. So to me it's a toss-up whether doing this adds value or not.

> There hasn't been any objection so far to delivering the target with a custom OSD configuration (one for DJI one for analog). Is that acceptable or does that also fall under user preference and need to be removed?

Personally as a user I wouldn't want the OSD elements predefined. But from a developer perspective I'm a bit more on the fence about this and don't have a strong opinion. At a high level I think it's generally a bad idea to "bake" these into the unified target config however. One reason is that it might be perceived to be a benefit to the user in that they don't have to set them up, they are however **forced to turn them off** if they don't want them. And will have to turn them off anytime the board is reset to defaults The other aspect is that future enhancements to the OSD item position definitions (the parameter values) are planned and what's stored in the unified config may not work as expected in the future. So in that case the user might get an error or unexpected result when applying defaults. This is actually the strongest argument against applying non-hardware definition settings in the unified config. We're aware of the dependence of unified targets on specific configuration settings and therefore are careful about how they're used, changed, etc. We won't be expecting user preference or other general settings in the unified config so this will eventually cause a problem.

configs/default/IFRC-JBF7.config
```diff
@@ -0,0 +1,212 @@
+# version
```
mikeller, line 1: The `# version` line needs to be removed.

configs/default/IFRC-JBF7_DJI.config
```diff
@@ -0,0 +1,175 @@
+# version
```
mikeller, line 1: Same here.

configs/default/IFRC-JBF7.config
Linjieqiang: Thank you for your reminder, it has been corrected.

configs/default/IFRC-JBF7_DJI.config
Linjieqiang: Thank you for your reminder, it has been corrected.

Linjieqiang:
Modify the configuration as required. Rebased and pushed it.

configs/default/IFRC-JBF7_DJI.config
Linjieqiang: Thank you for your reminder, it has been removed.

configs/default/IFRC-JBF7.config
Linjieqiang: In order to avoid legal risks, the VTX form has been removed.

configs/default/IFRC-JBF7.config
```diff
@@ -0,0 +1,192 @@
...
+set ibata_scale = 100
+set beeper_inversion = ON
+set beeper_od = OFF
+set sdcard_mode = SPI
+set sdcard_spi_bus = 3
+set osd_warn_batt_not_full = OFF
+set osd_warn_batt_warning = OFF
+set osd_warn_batt_critical = OFF
+set osd_warn_esc_fail = OFF
+set osd_warn_core_temp = OFF
```
etracer65, line 142: Disabling the warnings in the unified target config is a very bad idea. The decision whether or not a warning is important would be the user's decision.

configs/default/IFRC-JBF7.config
```diff
@@ -0,0 +1,192 @@
...
+set osd_rssi_alarm = 40
```
etracer65, line 143: I don't understand why this should be changed or be part of a unified target config. There's nothing about the flight controller that has any bearing on the receiver's RSSI.

configs/default/IFRC-JBF7.config
```diff
@@ -0,0 +1,192 @@
...
+set osd_cap_alarm = 5000
```
etracer65, line 144: Why would this be part of the unified target config? Does this flight controller only ship with and allow use with a 5000mah battery?

configs/default/IFRC-JBF7.config
```diff
@@ -0,0 +1,192 @@
...
+set mag_bustype = I2C
+set mag_i2c_device = 2
+set baro_bustype = I2C
+set baro_i2c_device = 2
+set serialrx_provider = SBUS
+set adc_device = 3
+set blackbox_p_ratio = 64
+set blackbox_device = SDCARD
+set dshot_burst = ON
+set vbat_max_cell_voltage = 440
```
etracer65, line 129: The max cell voltage setting has nothing to do with the flight controller and shouldn't be set in the target config.

configs/default/IFRC-JBF7_DJI.config
```diff
@@ -0,0 +1,163 @@
...
+set baro_bustype = I2C
+set baro_i2c_device = 2
+set min_check = 1010
+set max_check = 1990
+set serialrx_provider = SBUS
+set adc_device = 3
+set blackbox_p_ratio = 64
+set blackbox_device = SDCARD
+set dshot_burst = ON
+set vbat_max_cell_voltage = 440
```
etracer65, line 134: The max cell voltage setting has nothing to do with the flight controller and shouldn't be set in the target config.

configs/default/IFRC-JBF7_DJI.config
```diff
@@ -0,0 +1,163 @@
...
+serial 3 1024 115200 57600 0 115200
+
+# master
+set gyro_to_use = BOTH
+set mag_bustype = I2C
+set mag_i2c_device = 2
+set baro_bustype = I2C
+set baro_i2c_device = 2
+set min_check = 1010
+set max_check = 1990
```
etracer65, line 128: The `min_check` and `max_check` settings are user preferences based on the radio type and preferred "deadzones". They have nothing to do with the operation of the flight controller and shouldn't be in the target config.

configs/default/IFRC-JBF7.config
```diff
@@ -0,0 +1,192 @@
...
+
+# master
+set gyro_to_use = BOTH
+set mag_bustype = I2C
+set mag_i2c_device = 2
+set baro_bustype = I2C
+set baro_i2c_device = 2
+set serialrx_provider = SBUS
+set adc_device = 3
+set blackbox_p_ratio = 64
```
etracer65, line 126: I don't see why it's appropriate to change the blackbox logging rate to 2khz in the target config. What is it about this board that requires 2khz logging?

**Also this setting will not work with versions after 4.2 and will generate errors.** The way blackbox logging rates are set has been changed. This is another example as to why "preference" settings should never be in the unified target config.

configs/default/IFRC-JBF7_DJI.config
```diff
@@ -0,0 +1,163 @@
...
+set serialrx_provider = SBUS
+set adc_device = 3
+set blackbox_p_ratio = 64
```
etracer65, line 131: I don't see why it's appropriate to change the blackbox logging rate to 2khz in the target config. What is it about this board that requires 2khz logging?

**Also this setting will not work with versions after 4.2 and will generate errors.** The way blackbox logging rates are set has been changed. This is another example as to why "preference" settings should never be in the unified target config.

Linjieqiang:
@joshuabardwell

mikeller (approved):
Updated, should be ready now.

configs/default/IFRC-JBF7.config
etracer65: The OSD warnings are still disabled. I'm assuming this was just missed or maybe there's disagreement on this?

mikeller: As I'd said before I was ok with the proposed configurations. According to the policy in https://github.com/betaflight/betaflight/blob/master/docs/TargetMaintenance/CreatingAUnifiedTarget.md they are discouraged, but if manufacturers feel it is appropriate to use them then this is their decision. But apparently you want to implement your own policy.

PR #328, opened 2020-08-03
add AIRBOTF4V2

x4FF3:
Hi,

the target for https://github.com/betaflight/betaflight/issues/9994

KR
David

### DIFF
--- a/configs/default/AIRB-AIRBOTF4V2.config
+++ b/configs/default/AIRB-AIRBOTF4V2.config
@@ -1,0 +1,97 @@
+# Betaflight / STM32F405 (S405) 4.2.1 Jul 19 2020 / 06:18:01 (caa0d683c) MSP API: 1.43
+
+board_name AIRBOTF4V2
+manufacturer_id AIRB
+
+# resources
+resource BEEPER 1 C05
+resource MOTOR 1 C08
+resource MOTOR 2 B06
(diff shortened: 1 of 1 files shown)


configs/default/AIRB-AIRBOTF4V2.config
```diff
@@ -0,0 +1,99 @@
...
+# feature
+feature -RX_PARALLEL_PWM
+feature RX_SERIAL
+feature OSD
+
+# serial
+serial 0 64 115200 57600 0 115200
+
+# master
+set mag_hardware = NONE
```
mikeller, line 81: If this setting is used, this means that users will not be able to use any magnetometer on this board, and if they open support requests about this they will be told they will have to buy a different flight controller if they want to use a magnetometer.

configs/default/AIRB-AIRBOTF4V2.config
```diff
@@ -0,0 +1,99 @@
...
+set baro_hardware = NONE
```
mikeller, line 82: If this setting is used, this means that users will not be able to use any barometer on this board, and if they open support requests about this they will be told they will have to buy a different flight controller if they want to use a magnetometer.

x4FF3:
thanks for the hint. will fix it tonight.

PR #329, opened 2020-08-05
added ZEEZ-ZEEZWHOOP.config , small fix to ZEEZ-ZEEZF7.config

Guidus93:
-Added new Zeez Whoop U-Target
- Small FIX to Zeez F7 U-Target (added LED setting, fixed VBAT scale, added OSD core temp value)

### DIFF
--- a/configs/default/ZEEZ-ZEEZWHOOP.config
+++ b/configs/default/ZEEZ-ZEEZWHOOP.config
@@ -1,0 +1,84 @@
+# Betaflight / STM32F411 (S411) 4.1.0 Oct 16 2019 / 11:57:34 (c37a7c91a) MSP API: 1.42
+
+board_name ZEEZWHOOP
+manufacturer_id ZEEZ
+
+# resources
+resource MOTOR 1 A00
+resource MOTOR 2 A01
+resource MOTOR 3 B06
+resource MOTOR 4 B07
+resource PPM 1 A03
+resource LED_STRIP 1 A08
+resource SERIAL_TX 1 A09
+resource SERIAL_TX 2 A02
+resource SERIAL_RX 1 A10
+resource SERIAL_RX 2 A03
+resource I2C_SCL 1 B08
+resource I2C_SDA 1 B09
+resource LED 1 C13
+resource SPI_SCK 1 A05
+resource SPI_SCK 3 B03
+resource SPI_MISO 1 A06
+resource SPI_MISO 3 B04
+resource SPI_MOSI 1 A07
+resource SPI_MOSI 3 B05
+resource ADC_BATT 1 B01
+resource ADC_CURR 1 A04
+resource OSD_CS 1 B10
+resource GYRO_EXTI 1 C15
+resource GYRO_CS 1 A15
+
+# timer
+timer A03 AF3
+# pin A03: TIM9 CH2 (AF3)
+timer A00 AF1
+# pin A00: TIM2 CH1 (AF1)
+timer A01 AF1
+# pin A01: TIM2 CH2 (AF1)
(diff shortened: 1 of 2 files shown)


configs/default/ZEEZ-ZEEZWHOOP.config
```diff
@@ -0,0 +1,84 @@
...
+serial 1 64 115200 57600 0 115200
+
+# master
+set mag_hardware = NONE
+set baro_hardware = NONE
+set serialrx_provider = CRSF
+set blackbox_device = NONE
+set dshot_burst = AUTO
+set dshot_bitbang = OFF
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 75: If this setting is used as a default, and if your board is used in combination with older ESCs that do not support Dshot, this will result in the motors spinning up as soon as a battery is connected, potentially resulting in damage or injury.

configs/default/ZEEZ-ZEEZWHOOP.config
```diff
@@ -0,0 +1,84 @@
...
+set current_meter = NONE
```
mikeller, line 76: Please remove the `ADC_CURR` resource definition if this board is to not support a current meter.

configs/default/ZEEZ-ZEEZWHOOP.config
```diff
@@ -0,0 +1,84 @@
...
+# feature
+feature -RX_PARALLEL_PWM
+feature RX_SERIAL
+feature OSD
+
+# serial
+serial 1 64 115200 57600 0 115200
+
+# master
+set mag_hardware = NONE
```
mikeller, line 69: If this setting is used, this means that users will not be able to use any magnetometer on this board, and if they open support requests about this they will be told they will have to buy a different flight controller if they want to use a magnetometer.

configs/default/ZEEZ-ZEEZWHOOP.config
```diff
@@ -0,0 +1,84 @@
...
+set baro_hardware = NONE
```
mikeller, line 70: If this setting is used, this means that users will not be able to use any barometer on this board, and if they open support requests about this they will be told they will have to buy a different flight controller if they want to use a barometer.

configs/default/ZEEZ-ZEEZWHOOP.config
Guidus93: Hello Mike,
it's a tiny 1s AIO board with 4 ESC integrated . There'rent pads for external ESC connections. ESC are routed out from the MCU to the onboard 4in1 ESC.

configs/default/ZEEZ-ZEEZWHOOP.config
Guidus93: Its an F411 so no GPIO available to connect the MAG and no space for place it into the board. Its a FC designed for 1S Whoop.

configs/default/ZEEZ-ZEEZWHOOP.config
Guidus93: Its an F411 so no GPIO available to connect the BARO and no space for place it into the board. Its a FC designed for 1S Whoop.

configs/default/ZEEZ-ZEEZWHOOP.config
Guidus93: Ok i will remove the ADC_CURR definition

configs/default/ZEEZ-ZEEZWHOOP.config
mikeller: Thanks, makes sense.

PR #330, opened 2020-08-11
Added bus settings for barometer to TMOTORF4.

mikeller:
Fixes betaflight/betaflight#10090.

### DIFF
--- a/configs/default/TMTR-TMOTORF4.config
+++ b/configs/default/TMTR-TMOTORF4.config
@@ -103,6 +103,8 @@ feature OSD
 # master
 set mag_bustype = I2C
 set mag_i2c_device = 2
+set baro_bustype = I2C
+set baro_i2c_device = 2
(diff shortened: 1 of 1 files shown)


etracer65:
I'm not sure this is a good idea. I was going to do this after resolving the issue but after looking deeper I decided not to. The board has an unpopulated pad for the baro which I assume is connected via SPI. The manufacturer page indicates the board is available with baro (as an option). The assumption about being connected via SPI is based on the target config for the F7 variant having a BARO_CS resource defined and it's the same physical layout (with the unpopulated baro pad).

So if anything I think the problem is the F4 variant missing the SPI definitions for the baro rather than making it I2C. If a board did ship with the onboard baro (or a user soldered it on) having the config be I2C would be even more confusing.

mikeller:
Ok, makes sense. At first glance the target definition looks like this was an oversight.

PR #332, opened 2020-08-13
Add BETAFPVF722 Unified Targets.

betafpv-engineer:
Add BETAFPVF722 Unified Targets.


configs/default/BEFH-BETAFPVF722.config
```diff
@@ -0,0 +1,115 @@
...
+feature RX_SERIAL
+feature OSD
+
+# master
+
+set baro_spi_device = 3
+set serialrx_provider = SBUS
+set blackbox_device = SPIFLASH
+set dshot_idle_value = 450
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 103: Unless this is only going to be used for an all-in-one board with integrated ESCs, this default setting is a safety risk, as it will result in spinning motors on power up when used with ESCs that do not support Dshot.
Is this target going to be used only for an all-in-one board?

liyouluo: Although the first one is an all-in-one board, I can't guarantee that it will only be used for an all-in-one board, so I will remove set motor_pwm_protocol = DSHOT600.

betafpv-engineer: Although the first one is an all-in-one board, I can't guarantee that it will only be used for an all-in-one board, so Dshot600 can be removed.

PR #337, opened 2020-08-26
IFLIGHT_SUCCEX_E: Remove the USART3 define. Just for IIC2.

Linjieqiang:


configs/default/IFRC-IFLIGHT_SUCCEX_E_F4.config
```diff
@@ -13,12 +13,10 @@ resource PPM 1 A03
 resource LED_STRIP 1 B06
 resource SERIAL_TX 1 A09
 resource SERIAL_TX 2 A02
-resource SERIAL_TX 3 B10
```
mikeller, line 16: Out of interest, why is support for UART3 removed?

mikeller (approved):

configs/default/IFRC-IFLIGHT_SUCCEX_E_F4.config
Linjieqiang: Beacuse the PB10 PB11 define for UART3 and IIC2 together. In order not to cause misunderstanding, we choose to disable serial port 3.

PR #338, opened 2020-08-27
Add TALONF7DJIHD target

bnn1044:
Add Heli-Nation F7 HD target, new FC design to mate up with DJI HD system.

### DIFF
--- a/configs/default/HENA-TALONF7DJIHD.config
+++ b/configs/default/HENA-TALONF7DJIHD.config
@@ -124,12 +124,8 @@ serial 4 64 115200 57600 0 115200
 serial 5 1 115200 57600 0 115200
 
 # master
-set mag_hardware = NONE
-set baro_hardware = NONE
 set serialrx_provider = SBUS
 set blackbox_device = SPIFLASH
-set dshot_burst = ON
-set motor_pwm_protocol = DSHOT600
 set current_meter = ADC
 set battery_meter = ADC
 set vbat_scale = 160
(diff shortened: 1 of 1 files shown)


configs/default/HENA-TALONF7DJIHD.config
```diff
@@ -0,0 +1,146 @@
...
+serial 4 64 115200 57600 0 115200
+serial 5 1 115200 57600 0 115200
+
+# master
+set mag_hardware = NONE
+set baro_hardware = NONE
+set serialrx_provider = SBUS
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 132: Is this target only ever going to be sold as an all-in-one with integrated ESC?

configs/default/HENA-TALONF7DJIHD.config
```diff
@@ -0,0 +1,146 @@
...
+feature RX_SERIAL
+feature OSD
+
+# serial
+serial 4 64 115200 57600 0 115200
+serial 5 1 115200 57600 0 115200
+
+# master
+set mag_hardware = NONE
+set baro_hardware = NONE
```
mikeller, line 128: If `mag_hardware = NONE` and `baro_hardware = NONE` is kept in here then Betaflight will not be supporting addition of baro or mag to this hardware, and users will be told to buy different hardware if they want to use baro or mag.
If you don't want this please remove these lines.

mikeller (changes requested):

mikeller (approved):

PR #339, opened 2020-08-29
Add MATEKF722MINISE target

TheIsotopes:
Add [Mateksys F722-miniSE](http://www.mateksys.com/?portfolio=f722-minise) target

### DIFF
--- a/configs/default/MTKS-MATEKF722MINISE.config
+++ b/configs/default/MTKS-MATEKF722MINISE.config
@@ -1,0 +1,144 @@
+# Betaflight / STM32F7X2 (S7X2) 4.3.0 Aug 29 2020 / 04:23:51 (2bec8fa33) MSP API: 1.44
+
+board_name MATEKF722MINISE
+manufacturer_id MTKS
+
+# resources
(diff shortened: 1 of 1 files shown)


MATEKSYS:
This FC uses MatekF722mini unified target.  no need new target.

TheIsotopes:
@MATEKSYS Yes I know.
but this is cleaned up and with `set baro_i2c_address = 118` the i2c error is also eliminated.

MATEKSYS:
Baro address has been set to 0x76 (118) by default in this PR  https://github.com/betaflight/betaflight/pull/10129

TheIsotopes:
Unfortunately, an i2c error occurs anyway, which can only be corrected with `set baro_i2c_address = 118` as default.

MATEKSYS:
Normal. The errors occur during initialisation before the I2C speeds are all set up right.

Asizon:
Latest 4.3 development version should improve initialization and i2c detected errors

TheIsotopes:
ok then i close here.

PR #340, opened 2020-08-30
Fix camera switching to C2 by default on GEPRCF722

ItsMeAubey:
This board has its USER1 mode range set to 1400-2100, thus switching the camera to the C2 pad by default. Fixed by changing default to 1800-2100 - felt that correcting the default was more appropriate than removing the definition.


mikeller:
@Linjieqiang: Are you ok with this change, and does this match the documentation, or can the documentation be updated?

ItsMeAubey:
This is an important issue. @Linjieqiang

mikeller (approved):

PR #341, opened 2020-08-31
Removed no longer existing parameter 'gyro_sync_denom' from target configurations.

mikeller:
Fixes betaflight/betaflight#10152.

### DIFF
--- a/configs/default/AIRB-OMNIBUSF4NANOV7.config
+++ b/configs/default/AIRB-OMNIBUSF4NANOV7.config
@@ -78,7 +78,6 @@ feature OSD
 serial 0 64 115200 57600 0 115200
 
 # master
-set gyro_sync_denom = 2
 set serialrx_provider = SBUS
 set blackbox_device = SPIFLASH
 set dshot_burst = ON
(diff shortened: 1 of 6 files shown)


McGiverGim (approved):
About this kind of change... what happens if a user installs some old version of Betaflight? 4.1 I think supports unified config and has this parameter.
I think remember we don' t have a "version control" for unified config files.

flosean (approved):

mikeller:
@McGiverGim: If a user installs an old version of the firmware with one of these Unified Targets, then the custom gyro denominator setting will obviously not be applied any more, and they will have to set the denominator manually, like it is the case for all other targets.
We currently do not have a way to allow multiple different versions of a Unified Target configuration for the same target to co-exist for different versions of the firmware. There are a number of ways that this could be implemented, like embedding the different configurations into a JSON file together with some metadata, allowing multiple configuration files with names indicating a version range to co-exist, or moving from a file based approach to a DB based approach. Pretty much all of these solutions will come with a drawback of increased effort to maintain the system, and / or increased complexity for the target maintainers to create and maintain targets, so it will be trade-off between value and effort.

PR #345, opened 2020-09-08
Add new board DFR_F722_DUAL_HD to DFRA manufacturer

mStrangers:
Add new board:
manufacturer : DFRA 
Board name : DFR_F722_DUAL_HD 
Base firmware : STM32F7x2 
Config from 4.2.2

### DIFF
--- a/configs/default/DFRA-DFR_F722_DUAL_HD.config
+++ b/configs/default/DFRA-DFR_F722_DUAL_HD.config
@@ -118,7 +118,6 @@ set serialrx_provider = SBUS
 set adc_device = 3
 set blackbox_device = SPIFLASH
 set dshot_burst = ON
-set motor_pwm_protocol = DSHOT300
 set current_meter = ADC
 set battery_meter = ADC
 set ibata_scale = 100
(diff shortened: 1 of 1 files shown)


configs/default/DFRA-DFR_F722_DUAL_HD.config
```diff
@@ -0,0 +1,136 @@
...
+set gyro_to_use = BOTH
+set baro_bustype = I2C
+set baro_i2c_device = 2
+set mag_bustype = I2C
+set mag_i2c_device = 2
+set serialrx_provider = SBUS
+set adc_device = 3
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT300
```
mikeller, line 121: Will this only be sold as an all-in-one board with ESC or in a kit with ESC? If not this will have to be removed.

mikeller (changes requested):

configs/default/DFRA-DFR_F722_DUAL_HD.config
mStrangers: This one can be sold alone so i had removed the protocol declaration

mikeller (approved):

PR #359, opened 2020-10-02
DroniusF7-Fixed non-usb-powered MAX7456

Asizon:
Another target with broken osd chip definition.


mikeller:
@cvetaevvitaliy: As the creator of this target, can you confirm that the MAX7456 in this design is only powered from the battery input but not from USB?

cvetaevvitaliy:
Hi! 
Yes, OSD MAX7456 chip is powered by battery, for guaranteed 5V supply when downloading and save fonts in OSD chip.

mikeller (approved):

PR #366, opened 2020-10-22
Update DIAT-MAMBAF722.config

kc10kevin:
Diatone requesting change MAMBAF722 back to orginal.  They made a mistake when making the change.  Sorry!  Thanks Michael.

### DIFF
--- a/configs/default/DIAT-MAMBAF722.config
+++ b/configs/default/DIAT-MAMBAF722.config
@@ -100,8 +100,6 @@ serial 0 64 115200 57600 0 115200
 serial 3 1 19200 57600 0 115200
 
 # EXTRAS
-resource USB_DETECT 1 C13
-resource ESCSERIAL 1 B09
 resource PINIO 1 B00
 
 # feature
--- a/configs/default/DIAT-MAMBAF722.config
+++ b/configs/default/DIAT-MAMBAF722.config
@@ -97,11 +97,15 @@ dma pin B03 0  # pin B03: DMA1 Stream 6 Channel 3
 
 # serial
 serial 0 64 115200 57600 0 115200
+serial 3 1 19200 57600 0 115200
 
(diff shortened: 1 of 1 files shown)


configs/default/DIAT-MAMBAF722.config
```diff
@@ -114,5 +116,7 @@ set serialrx_provider = SBUS
 set blackbox_device = SPIFLASH
 set dshot_burst = ON
 set motor_pwm_protocol = DSHOT600
```
mikeller, line 118: To my knowledge this is sold as a dedicated flight controller without on-board ESCs - can you please remove this line to make it compliant with https://github.com/betaflight/betaflight/blob/master/docs/TargetMaintenance/CreatingAUnifiedTarget.md#13-donts?

kc10kevin: Waiting to hear back from Diatone.  ESCs not built into board, but pretty sure 90%+ of sales are complete rtf units with their own escs that have DHOT600 on default.  I believe that is why they want that in the config.  It’s been in the MAMBAF722 config since inception, so didn’t think it would be a problem keeping it in.  Your completely understand why it should not be there.   If they do not have a reason that is satisfactory to you then we will of course take it out.  Will comment when I hear from them.

kc10kevin: @mikeller Heard back from Diatone.  Exactly like I thought above.   They would like to keep DHOT600 in the config file, but understand it does not follow the rules because the board does not have build in ESCs supporting DSHOT.   There is no other argument to keep it in except the justification above, so if it must be removed then just say so and it's gone.  Code is waiting along with the name changes in the other pull request to push to you.  Thanks again and sorry for the issues.

mikeller (approved):

PR #369, opened 2020-10-24
Add RUSH-BLADE_F7_HD flight controller support.

DusKing1:
We use a plug-in socket for DJI's connector and provide two types of wires, so if the user decides not to use DJI air unit's build-in S.Bus line, they can use UART1 or UART5, and we'll take the risk of this, and will remind users on our website and in manual.

### DIFF
--- a/configs/default/RUSH-BLADE_F7_HD.config
+++ b/configs/default/RUSH-BLADE_F7_HD.config
@@ -1,8 +1,4 @@
-# version
-# Betaflight / RUSHFCHD (RF7H) 4.2.1 Aug 13 2020 / 00:11:54 (caa0d683c) MSP API: 1.43
-
-# start the command batch
-batch start
+# Betaflight / STM32F7X2 (S7X2) 4.2.4 Oct 20 2020 / 08:20:06 (fbcaf8c50) MSP API: 1.43
 
 board_name BLADE_F7_HD
 manufacturer_id RUSH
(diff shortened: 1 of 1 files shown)


configs/default/RUSH-BLADE_F7_HD.config
```diff
@@ -0,0 +1,113 @@
+# version
+# Betaflight / RUSHFCHD (RF7H) 4.2.1 Aug 13 2020 / 00:11:54 (caa0d683c) MSP API: 1.43
```
mikeller, line 2: This will not work:
- the `# Betaflight...` line needs to be the first one in the file;
- this line needs to be for the appropriate 'STM32...' unified target, or else configurator will not know which target to flash.

configs/default/RUSH-BLADE_F7_HD.config
```diff
@@ -0,0 +1,113 @@
...
+
+# start the command batch
+batch start
```
mikeller, line 5: This should not have `batch` commands in it.

mikeller (changes requested):

DusKing1:
Sorry, forgot to modified those commands

configs/default/RUSH-BLADE_F7_HD.config
DusKing1: oh forgot to change this

configs/default/RUSH-BLADE_F7_HD.config
DusKing1: This is my negligence, sorry!

mikeller (approved):

PR #367, opened 2020-10-23
ADD  VGRC-VGOODRCF411_DJI  targets config

hljxkang2020:
Hi Mikeller,
For the  safety about default with Dshot 600:
We sell this controller with our latest BLHeli_32 ESC and our ESC default also with Dshot600, at the same time, we will write and enhance this safety instruction in our FC manual. We will take responsibility for any safety issue if there. tks.

PR #372, opened 2020-10-27
ADD  VGRC-VGOODRCF411_DJI  targets config

hljxkang2020:
Hi Mikeller,
For the  safety about default with Dshot 600:
We sell this controller with our latest BLHeli_32 ESC and our ESC default also with Dshot600, at the same time, we will write and enhance this safety instruction in our FC manual. We will take responsibility for any safety issue if there. tks.


mikeller:
Duplicate of #367. Please re-open and make the change on your existing pull request.

PR #375, opened 2020-10-29
Add EACH-EACHINEF722_AIO

EachineRC:
Please add this target for EACH.

### DIFF
--- a/configs/default/EACH-EACHINEF722_AIO.config
+++ b/configs/default/EACH-EACHINEF722_AIO.config
@@ -1,0 +1,131 @@
+# Betaflight / STM32F7X2 (S7X2) 4.1.3 Jan 16 2020 / 11:35:53 (543a5e1ba) MSP API: 1.42
+
+board_name EACHINEF722_AIO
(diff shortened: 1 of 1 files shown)


configs/default/EACH-EACHINEF722_AIO.config
```diff
@@ -0,0 +1,131 @@
...
+# serial
+serial 0 64 115200 57600 0 115200
+
+# master
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set baro_spi_device = 1
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 119: Am I right to assume that this is an AIO board with on-board ESCs that are capable of Dshot600?

EachineRC: Yes, EACHINE F722 AIO is based on F722 FC integrated BLHELI32 ESC.

mikeller (approved):

PR #374, opened 2020-10-28
updata VGRC-VGOODRCF411_DJI config

hljxkang2020:

PR #373, opened 2020-10-28
ADD  VGRC-VGOODRCF722_DJI_MINI targets

hljxkang2020:
For the  safety about default with Dshot 600:
We sell this controller with our latest BLHeli_32 ESC and our ESC default also with Dshot600, at the same time, we will write and enhance this safety instruction in our FC manual. We will take responsibility for any safety issue if there. tks.

PR #379, opened 2020-11-03
Add VGOODRC F4 and F7 FC

flosean:
Help VGOODRC to add F4 and F7 flight controller


mikeller:
Duplicate of #374, #373, #372, and #367. 
As stated in https://github.com/betaflight/unified-targets/pull/374#issuecomment-717885916, dealing with manufacturers who are too cheap and don't care enough about their customers to hire somebody capable of maintaining their targets is a waste.
No further targets or from this manufacturer will be accepted.

PR #382, opened 2020-11-04
Changed maintainer for AIKON, Spedix, T-motor.

mikeller:
@DieHertz: I had a chat with @Guidus93, and he is now taking care of the maintenance of targets for AIKON, Spedix, and T-motor. Please let me know if you would like to stay on as a maintainer, or if you should be removed.


DieHertz:
Hi, better remove me then

PR #384, opened 2020-11-07
Add TCMMF411

hongzhou-zeng:
Hi Mikeller,
For the safety about default with Dshot 600:
we will write and enhance this safety instruction in FC manual. We will take responsibility for any safety issue if there. thanks!


配置/默认/TCMMF411
```diff
@@ -0,0 +1,112 @@
+# Betaflight / STM32F411 (S411) 4.1.1 Nov 15 2019 / 12:55:17 (1e5e3d369) MSP API: 1.42
```
mikeller, line 1: Path and file name for this are wrong - it needs to be in `configs/default/<manufacturer id>-<board name>.config

配置/默认/TCMMF411
```diff
@@ -0,0 +1,112 @@
...
+# pin B03: DMA1 Stream 6 Channel 3
+dma pin B10 0
+# pin B10: DMA1 Stream 1 Channel 3
+dma pin A08 0
+# pin A08: DMA2 Stream 6 Channel 0
+
+# feature
+feature -RX_PARALLEL_PWM
+feature RX_SERIAL
+feature SOFTSERIAL
```
mikeller, line 78: Why are you enabling SOFTSERIAL if there are no pins defined for this function?

配置/默认/TCMMF411
```diff
@@ -0,0 +1,112 @@
...
+led 0 0,0::A:2
+led 1 1,0::A:2
+led 2 2,0::A:2
+led 3 3,0::A:2
+
+# master
+set baro_spi_device = 2
+set serialrx_provider = SBUS
+set blackbox_device = SPIFLASH
+set dshot_idle_value = 450
```
mikeller, line 96: Will this target only be used for a ready to fly quad that comes with ESCs / motors / props pre-installed? If not then drivetrain specific settings should not be part of the configuration.

配置/默认/TCMMF411
```diff
@@ -0,0 +1,112 @@
...
+set dshot_burst = AUTO
+set dshot_bitbang = OFF
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 99: Unless this target is only used for an AIO board with integrated ESCs, this can not be set.

配置/默认/TCMMF411
```diff
@@ -0,0 +1,112 @@
...
+set current_meter = ADC
+set battery_meter = ADC
+set mag_hardware = NONE
```
mikeller, line 102: If this is set, users wanting to add a compass to their board will be told to buy hardware from a different manufacturer - please leave this on the default if you do not want to lose customers.

配置/默认/TCMMF411
```diff
@@ -0,0 +1,112 @@
...
+set beeper_inversion = ON
+set beeper_od = OFF
+set system_hse_mhz = 8
```
mikeller, line 105: Do not explicitly set default values.

mikeller (changes requested):

配置/默认/TCMMF411
hongzhou-zeng: I had deleted SOFTSERIAL

配置/默认/TCMMF411
hongzhou-zeng: Yes,this target is only used for TCMMF411 AIO board

配置/默认/TCMMF411
hongzhou-zeng: Yes，I had deleted

配置/默认/TCMMF411
hongzhou-zeng: yes,I had deleted 。

hongzhou-zeng:
@mikeller    I had updated the TCMMF411 config file,Please check it is right?thanks

PR #386, opened 2020-11-11
add AOCODAF722MINI

star624:
hello,

please help to add new target for SJET


configs/default/SJET-AOCODAF722MINI.config
```diff
@@ -0,0 +1,146 @@
...
+feature -RX_PARALLEL_PWM
+feature RX_SERIAL
+feature OSD
+
+# serial
+serial 0 64 115200 57600 0 115200
+
+# master
+set serialrx_provider = SBUS
+set align_mag = DEFAULT
```
mikeller, line 101: Please remove all of the default values from this.

mikeller (changes requested):

configs/default/SJET-AOCODAF722MINI.config
star624: ok,  i have remove it and the declarations of gyro_1.

mikeller (approved):
Much better. :-)

PR #390, opened 2020-11-15
Add TCMMF411.config

ThanYangFPV:
This is a new start. 
Please ignore pull requests that were wrong before.
Please do not add flags that are duplicate.
This configuration is for AIO use only.Dshot600 is enabled by default.

thank you.

### DIFF
--- a/configs/default/TCMM-TCMMF411.config
+++ b/configs/default/TCMM-TCMMF411.config
@@ -91,7 +91,6 @@ led 3 3,0::A:2
 set baro_spi_device = 2
 set serialrx_provider = SBUS
 set blackbox_device = SPIFLASH
-set dshot_idle_value = 450
 set dshot_burst = AUTO
 set dshot_bitbang = OFF
 set motor_pwm_protocol = DSHOT600
--- a/configs/default/TCMM-TCMMF411.config
+++ b/configs/default/TCMM-TCMMF411.config
@@ -77,7 +77,6 @@ feature -RX_PARALLEL_PWM
 feature RX_SERIAL
 feature OSD
 feature LED_STRIP
-feature AIRMODE
 
 # serial
 serial 0 64 115200 57600 0 115200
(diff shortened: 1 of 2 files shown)


configs/default/TCMMF411.config
```diff
@@ -0,0 +1,110 @@
+# Betaflight / STM32F411 (S411) 4.1.1 Nov 15 2019 / 12:55:17 (1e5e3d369) MSP API: 1.42
+
+board_name TCMMF411
```
mikeller, line 3: The file name for this is wrong - please read and follow the [instructions](https://github.com/betaflight/betaflight/blob/master/docs/TargetMaintenance/CreatingAUnifiedTarget.md).

configs/default/TCMMF411.config
```diff
@@ -0,0 +1,110 @@
...
+# pin B10: DMA1 Stream 1 Channel 3
+dma pin A08 0
+# pin A08: DMA2 Stream 6 Channel 0
+
+# feature
+feature -RX_PARALLEL_PWM
+feature RX_SERIAL
+feature OSD
+feature LED_STRIP
+feature AIRMODE
```
mikeller, line 80: Please follow the instructions - explicitly adding default settings is not acceptable.

mikeller (changes requested):

configs/default/TCMMF411.config
ThanYangFPV: OK.

configs/default/TCMMF411.config
ThanYangFPV: OK,

configs/default/TCMM-TCMMF411.config
```diff
@@ -0,0 +1,109 @@
...
+led 0 0,0::A:2
+led 1 1,0::A:2
+led 2 2,0::A:2
+led 3 3,0::A:2
+
+# master
+set baro_spi_device = 2
+set serialrx_provider = SBUS
+set blackbox_device = SPIFLASH
+set dshot_idle_value = 450
```
etracer65, line 94: This is a very bad idea. Setting the idle too low can cause motor de-syncs and serious support issues.

ThanYangFPV: Good suggestion, use the default idle value.

mikeller (approved):

PR #393, opened 2020-11-19
Added manufacturer id and maintainer for Mayhem RC.

mikeller:
Fixes https://github.com/betaflight/betaflight/issues/10354


JyeSmith (commented):
Accidental deletion of SJET-AOCODAF722MINI.config in this PR?

mikeller:
@JyeSmith: Thanks for noticing - that's the one that had MS-DOS line endings and had to be fixed in #394.

PR #396, opened 2020-11-20
BARO_CS actually USES C05

ThanYangFPV:
Hi, everybody.
I will Replace hongzhou-zeng to maintain all objectives.

For this firmware, in our flight control on shipment,BARO_CS actually USES C05.

Thank you.


mikeller (approved):

mikeller:
@ThanYangFPV:

> I will Replace hongzhou-zeng to maintain all objectives.

Does this apply to all manufacturers that he is responsible for (BEFH / SKST / SNMT)?

@hongzhou-zeng: Can you confirm this?

PR #399, opened 2020-11-23
Diatone Baro/Mag to AUTO

kc10kevin:
Diatone Baro/Mag to AUTO

Thanks Michael.   Sorry for all the problems.

### DIFF
--- a/configs/default/DIAT-FURYF4.config
+++ b/configs/default/DIAT-FURYF4.config
@@ -78,7 +78,6 @@ dma pin A00 0
 # pin A00: DMA1 Stream 2 Channel 6
 
 # master
-set baro_hardware = AUTO
 set mag_bustype = I2C
 set mag_i2c_device = 1
 set baro_bustype = I2C
--- a/configs/default/DIAT-FURYF4.config
+++ b/configs/default/DIAT-FURYF4.config
@@ -78,7 +78,6 @@ dma pin A00 0
 # pin A00: DMA1 Stream 2 Channel 6
 
 # master
-set mag_hardware = AUTO
 set baro_hardware = AUTO
 set mag_bustype = I2C
 set mag_i2c_device = 1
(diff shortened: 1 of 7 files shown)


configs/default/DIAT-MAMBAF722_I2C.config
```diff
@@ -120,7 +120,7 @@ feature OSD
 # master
 set baro_bustype = I2C
 set baro_i2c_device = 1
-set mag_hardware = NONE
+set mag_hardware = AUTO
```
mikeller, line 123: `AUTO` is the default for all of these - they should just be omitted, which removes the risk of them getting stale when a parameter is renamed or dropped.

kc10kevin: Thanks.   Removed.

configs/default/DIAT-FURYF4.config
```diff
@@ -78,6 +78,7 @@ dma pin A00 0
 # pin A00: DMA1 Stream 2 Channel 6
 
 # master
+set baro_hardware = AUTO
```
mikeller, line 81: Sorry, looks like I was not explicit enough - `baro_hardware = AUTO` is a default as well and should not be explicitly used in a Unified Target configuration.

Hint: You can check this yourself, the default is listed for parameters on `get` if they are not already set to the default:

```
# get mag_hardware
mag_hardware = NONE
Allowed values: AUTO, NONE, HMC5883, AK8975, AK8963, QMC5883, LIS3MDL, MAG_MPU925X_AK8963
Default value: AUTO
```

kc10kevin: Sorry about that.  Misunderstood what you were getting at.   Got it now.   I’ll remove completely and push the change once I get back from vacation.  Thanks again.

kc10kevin: @mikeller Complete.  All removed.  Thanks again.

mikeller (approved):

PR #400, opened 2020-11-26
add HARC-HAKRCF405D.config

WilsonChang:
Hi,
please review and add HARC-HAKRCF405D unified configure target file for HAKRC.
thanks.

### DIFF
--- a/configs/default/HARC-HAKRCF405D.config
+++ b/configs/default/HARC-HAKRCF405D.config
@@ -116,13 +116,8 @@ serial 1 64 115200 57600 0 115200
 # master
 set baro_bustype = I2C
 set baro_i2c_device = 1
-set mag_bustype = I2C
-set mag_i2c_device = 1
-set mag_hardware = NONE
 set serialrx_provider = SBUS
 set blackbox_device = SPIFLASH
-set dshot_burst = ON
-set motor_pwm_protocol = DSHOT600
 set current_meter = ADC
 set battery_meter = ADC
 set beeper_inversion = ON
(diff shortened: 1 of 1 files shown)


configs/default/HARC-HAKRCF405D.config
```diff
@@ -0,0 +1,140 @@
...
+
+# serial
+serial 1 64 115200 57600 0 115200
+
+# master
+set baro_bustype = I2C
+set baro_i2c_device = 1
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set mag_hardware = NONE
```
mikeller, line 121: If this is set, this means that mag devices will not be supported on this flight controller, and users asking for instructions will be told to buy a different flight controller.
If you do not want this, please remove this line - the default setting works just fine when no mag device is connected.

configs/default/HARC-HAKRCF405D.config
```diff
@@ -0,0 +1,140 @@
...
+set serialrx_provider = SBUS
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 125: Is this target only going to be used for an all-in-one board with integrated ESCs that support Dshot600?

mikeller (changes requested):

configs/default/HARC-HAKRCF405D.config
WilsonChang: OK!

configs/default/HARC-HAKRCF405D.config
WilsonChang: Only a suggestion. OK I delete this.

mikeller (approved):

PR #402, opened 2020-11-28
LDARC_F411: change something

Linjieqiang:


configs/default/LDAR-LDARC_F411.config
```diff
@@ -8,14 +9,14 @@ resource MOTOR 1 B04
 resource MOTOR 2 B05
 resource MOTOR 3 B06
 resource MOTOR 4 B07
-resource MOTOR 5 B03
```
mikeller, line 11: You can actually leave those ones in - the pins will then be available either as softserial or as motor pins.

Linjieqiang: Forgive me for being so late. For this modification, I initiated the pull request in accordance with the document given to me by the LDAR company, because they have already tested it, so I did not review it again. I think it would be fine to merge directly.

mikeller (approved):

PR #403, opened 2020-12-05
add speedybeef7v2 unified target

azolyoung:
Hello,

We are going to release a new F7 flight controller, the following info is the main spec:
MCU: F722
GYRO: MPU6000
FLASH: 16MB flash
ESC POWER SUPPLY: NO 
UART: 5 FULL UART and 1 UART RX for ESC Telemetry.
BEC: 5V & 9V BEC

it has some recommend-wiring\inner connected with it so we'd like to add some pre-settings in the config, not sure if these are appropriate:

#### 1. Use SerialRx and SBUS as default receiver settings, in our recommended wiring diagram, we use the serial receiver and UART1 RX to demonstrate the receiver wiring, so set a preset  RX setting:
```
serial 0 64 115200 57600 0 115200
...
feature -RX_PARALLEL_PWM
feature RX_SERIAL
set serialrx_provider = SBUS
```

#### 2. The FC has an 8 pins JST slot,  there is a TEL pin in it, it connects to RX6 of MCU,  so we'd like to enable the ESC SENSOR on UART6
```
feature ESC_SENSOR
...
serial 5 1024 115200 57600 0 115200
```
#### 3. We noticed set a preset ESC protocol is not recommended in the instructions, we understand it's a safety consideration, but in our market plan, this FC will sell with our BLHeli32 45A ESC as a combo(the customer still can buy FC or ESC alone), we'd like to know if is acceptable to set 'DSHOT600' as our default ESC protocol setting? We will add the safety notice on the product page.
```
set motor_pwm_protocol = DSHOT600
```

Hello,

We are going to release a new F7 flight controller, the following info is the main spec:
MCU: F722
GYRO: MPU6000
FLASH: 16MB flash
ESC POWER SUPPLY: NO
UART: 5 FULL UART and 1 UART RX for ESC Telemetry.
BEC: 5V & 9V BEC

There are some presets in the firmware defs:
1. This FC will have a combo with our BLHELI32 45A ESC, so we add some preset for the ESC, the presets are include
a. enable ESC_SENSOR
b. use DSHOT600 as default ESC protocol
c. turn on the "ESC" function on the UART4
2. Use SBUS as our default RX protocol, and in our recommended wiring diagram is use SBUS as our default RX device. 

Thanks for your review!

### DIFF
--- a/configs/default/SPBE-SPEEDYBEEF7V2.config
+++ b/configs/default/SPBE-SPEEDYBEEF7V2.config
@@ -1,0 +1,113 @@
+# Betaflight / STM32F7X2 (S7X2) 4.2.5 Nov 22 2020 / 18:39:53 (afdac08b3) MSP API: 1.43
+
+board_name SPEEDYBEEF7V2
+manufacturer_id SPBE
+
+# resources
+resource BEEPER 1 B02
+resource MOTOR 1 A08
+resource MOTOR 2 A09
+resource MOTOR 3 A10
+resource MOTOR 4 C09
+resource PPM 1 B07
+resource LED_STRIP 1 B00
+resource SERIAL_TX 1 B06
+resource SERIAL_TX 2 A02
+resource SERIAL_TX 3 C10
+resource SERIAL_TX 5 C12
+resource SERIAL_TX 6 C06
+resource SERIAL_RX 1 B07
+resource SERIAL_RX 2 A03
+resource SERIAL_RX 3 C11
+resource SERIAL_RX 4 A01
+resource SERIAL_RX 5 D02
+resource SERIAL_RX 6 C07
+resource I2C_SCL 1 B08
+resource I2C_SCL 2 B10
+resource I2C_SDA 1 B09
+resource I2C_SDA 2 B11
+resource LED 1 C13
+resource SPI_SCK 1 A05
+resource SPI_SCK 2 B13
+resource SPI_SCK 3 B03
+resource SPI_MISO 1 A06
+resource SPI_MISO 2 B14
+resource SPI_MISO 3 B04
+resource SPI_MOSI 1 A07
+resource SPI_MOSI 2 B15
+resource SPI_MOSI 3 B05
+resource CAMERA_CONTROL 1 A00
+resource ADC_BATT 1 C02
+resource ADC_RSSI 1 C03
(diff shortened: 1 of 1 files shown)


configs/default/SPBE-SPEEDYBEEF7V2.config
```diff
@@ -0,0 +1,134 @@
...
+set beeper_inversion = ON
+set beeper_od = OFF
+set max7456_spi_bus = 2
+set pinio_config = 129,1,1,1
+set pinio_box = 0,255,255,255
+set flash_spi_bus = 3
+set gyro_1_bustype = SPI
+set gyro_1_spibus = 1
+
+profile 0
```
mikeller, line 116: These profile changes here are not needed and should be removed.

configs/default/SPBE-SPEEDYBEEF7V2.config
```diff
@@ -0,0 +1,134 @@
...
+
+# master
+set mag_bustype = I2C
+set mag_i2c_device = 2
+set baro_bustype = I2C
+set baro_i2c_device = 1
+set serialrx_provider = SBUS
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 103: Unlike the other proposed custom defaults, this one will lead to the risk of unintended motor spin up if the fligth controller is used with certain ESCs. For this reason it is only acceptable if it is for an all-in-one design that does not allow the user to use non-original ESCs in an easy way.
Furthermore, if this is left at the default, the design of Betaflight configurator is such that the user is prompted to select an appropriate ESC protocol on connect, making this a good user experience, and not likely to result in support requests:

image

mikeller (changes requested):

configs/default/SPBE-SPEEDYBEEF7V2.config
azolyoung: OK! there waere removed.

configs/default/SPBE-SPEEDYBEEF7V2.config
azolyoung: Got it, We'll add a tip to tell the customer to choose the esc protocol manually.

mikeller (approved):

PR #404, opened 2020-12-12
FLYWOOF745NANO targets config

flywoo:
FLYWOOF745NANO targets config


configs/default/FLWO-FLYWOOF745NANO.config
```diff
@@ -0,0 +1,129 @@
...
+set battery_meter = ADC
+set beeper_inversion = ON
+set beeper_od = OFF
+set max7456_spi_bus = 2
+set dashboard_i2c_bus = 1
+set flash_spi_bus = 1
+set gyro_1_bustype = SPI
+set gyro_1_spibus = 4
+set gyro_1_sensor_align = CW270
+set gyro_2_spibus = 4
```
mikeller, line 129: This setting does not make sense, as gyro 2 is not configured.

mikeller (changes requested):

PR #406, opened 2020-12-17
Add VGRC-VGOODRCF722_DJI targets

hljxkang2020:
Add VGRC-VGOODRCF722_DJI targets

### DIFF
--- a/配置/默认/VGRC-VGOODRCF722_DJI
+++ b/配置/默认/VGRC-VGOODRCF722_DJI
@@ -1,0 +1,103 @@
+# Betaflight / STM32F7X2 (S7X2) 4.1.7 May 28 2020 / 15:06:20 (9ba02a587) MSP API: 1.42
+
+board_name VGOODRCF722_DJI
+manufacturer_id VGRC
+
+# resources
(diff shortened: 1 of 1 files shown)


配置/默认/VGRC-VGOODRCF722_DJI
```diff
@@ -0,0 +1,103 @@
...
+
+# serial
+serial 0 64 115200 57600 0 115200
+
+# master
+set serialrx_provider = SBUS
+set adc_device = 3
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 92: Is this an all-in-one offering that has the ESCs on board? If not this may not be used.

mikeller (changes requested):

配置/默认/VGRC-VGOODRCF722_DJI
hljxkang2020: Thanks for your remind! Yes, this is an all-in-one item.

mikeller (approved):

PR #407, opened 2020-12-17
Added TMTR_TMVeloxF7_config

tobeornottobe81:

### DIFF
--- a/configs/default/TMTR-TMVeloxF7.config
+++ b/configs/default/TMTR-TMVeloxF7.config
@@ -1,0 +1,144 @@
+# Betaflight / STM32F7X2 (S7X2) 4.2.5 Nov 22 2020 / 18:39:53 (afdac08b3) MSP API: 1.43
+board_name TMVeloxF7
+manufacturer_id TMTR
+
+# resources
+resource BEEPER 1 C14
+resource MOTOR 1 B01
+resource MOTOR 2 B04
+resource MOTOR 3 B03
+resource MOTOR 4 A15
+resource MOTOR 5 C08
+resource MOTOR 6 C09
+resource PPM 1 C06
+resource LED_STRIP 1 A08
+resource SERIAL_TX 1 A09
+resource SERIAL_TX 2 A02
+resource SERIAL_TX 3 B10
+resource SERIAL_TX 4 A00
+resource SERIAL_TX 5 C12
+resource SERIAL_TX 6 C06
+resource SERIAL_RX 1 A10
+resource SERIAL_RX 2 A03
+resource SERIAL_RX 3 B11
+resource SERIAL_RX 4 A01
+resource SERIAL_RX 5 D02
+resource SERIAL_RX 6 C07
+resource I2C_SCL 1 B06
+resource I2C_SDA 1 B07
+resource LED 1 C15
+resource SPI_SCK 1 A05
+resource SPI_SCK 2 B13
+resource SPI_SCK 3 C10
+resource SPI_MISO 1 A06
+resource SPI_MISO 2 B14
+resource SPI_MISO 3 C11
+resource SPI_MOSI 1 A07
+resource SPI_MOSI 2 B15
+resource SPI_MOSI 3 B05
+resource CAMERA_CONTROL 1 B08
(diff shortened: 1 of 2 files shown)


configs/default/TMTR-TMVeloxF7.config
```diff
@@ -0,0 +1,144 @@
+# Betaflight / STM32F7X2 (S7X2) 4.2.5 Nov 22 2020 / 18:39:53 (afdac08b3) MSP API: 1.43
+board_name TMVeloxF7
```
mikeller, line 2: There are places like OSD that cannot display lower case characters. Please use only upper case for board names.

configs/default/TMTR-TMVeloxF7.config
```diff
@@ -0,0 +1,144 @@
...
+dma pin A02 0
+# pin A02: DMA1 Stream 0 Channel 6
+dma pin A03 1
+# pin A03: DMA1 Stream 3 Channel 6
+
+# feature
+feature -RX_PARALLEL_PWM
+feature RX_SERIAL
+feature TELEMETRY
+feature RSSI_ADC
```
mikeller, line 110: Does this board come with an on-board RX that uses this outdated way to receive an RSSI value?
If not, this should not be pre-configured.

configs/default/TMTR-TMVeloxF7.config
```diff
@@ -0,0 +1,144 @@
...
+
+# master
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set baro_bustype = I2C
+set baro_i2c_device = 1
+set adc_device = 3
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 126: Is this an all-in-one offering that has the ESCs on board? If not this may not be used.

mikeller (changes requested):

configs/default/TMTR-TMVeloxF7.config
tobeornottobe81: OK, I will correct it

configs/default/TMTR-TMVeloxF7.config
tobeornottobe81: Yes, the target board has RSSI function

mikeller: What type of RX is on board on this flight controller - there probably are better way to get RSSI information than wasting an ADC input on an inaccurate way to transmit this information.

tobeornottobe81: Analog RSSI signal from receiver, ADC is idle when not in use

configs/default/TMTR-TMVeloxF7.config
tobeornottobe81: Yes, all of our ESC support dshot600

configs/default/TMTR-TMVeloxF7.config
mikeller: Why do you want to pre-select an outdated and poorly working option that is only needed for deprecated RX models, unless you had such a deprecated RX as part of your design? All current RX models support better ways to report RSSI.

configs/default/TMTR-TMVeloxF7.config
mikeller: So the user cannot connect this flight controller to a non - TMotor ESC. Can you elaborate a bit on how you achieve this?

configs/default/TMTR-TMVeloxF7.config
tobeornottobe81: OK, we will change the design and remove the RSSI_ ADC

When F3 flight control was popular, our OSD adopted atmega328 + at7456 scheme, but the original function was retained here. Now it doesn't seem necessary

configs/default/TMTR-TMVeloxF7.config
tobeornottobe81: At present, market ESC support PWM, oneshot125,dshot protocol and so on, and can automatically identify.

mikeller (approved):

PR #410, opened 2020-12-24
ADD GEPR-GEPRCF722BT targets

arbou:
ADD GRPR-GEPRCF722BT targets

### DIFF
--- a/configs/default/GEPR-GEPRCF722BT.config
+++ b/configs/default/GEPR-GEPRCF722BT.config
@@ -75,6 +75,7 @@ dma pin B03 0
 feature -RX_PARALLEL_PWM
 feature RX_SERIAL
 feature OSD
+feature gps
 
 # serial
 serial 1 64 115200 57600 0 115200
--- a/configs/default/GEPR-GEPRCF722BT.config
+++ b/configs/default/GEPR-GEPRCF722BT.config
@@ -1,0 +1,99 @@
+# Betaflight / STM32F7X2 (S7X2) 4.1.7 May 28 2020 / 15:06:20 (9ba02a587) MSP API: 1.42
+board_name GEPRCF722BT
+manufacturer_id GEPR
+
+# resources
+resource BEEPER 1 B02
+resource MOTOR 1 A09
+resource MOTOR 2 A08
+resource MOTOR 3 C09
+resource MOTOR 4 C08
+resource LED_STRIP 1 B03
+resource SERIAL_TX 1 B06
+resource SERIAL_TX 2 A02
+resource SERIAL_TX 3 B10
+resource SERIAL_TX 4 A00
+resource SERIAL_TX 5 C12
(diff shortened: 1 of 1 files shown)


configs/default/GEPR-GEPRCF722BT.config
```diff
@@ -0,0 +1,99 @@
...
+set mag_i2c_device = 2
+set baro_bustype = I2C
+set baro_i2c_device = 1
+set serialrx_provider = SBUS
+set adc_device = 3
+set blackbox_device = SPIFLASH
+set current_meter = ADC
+set battery_meter = ADC
+set beeper_inversion = ON
+set gps_provider = UBLOX
```
mikeller, line 95: Does this board come with an on-board GPS sensor? If not, there is not much point in pre-configuring the type of GPS sensor that the user will (or will not) connect.

arbou:
This is a muliti-in-one, it is with an on-board GPS sensor

configs/default/GEPR-GEPRCF722BT.config
arbou: This is a muliti-in-one, it is with an on-board GPS sensor

mikeller: Interesting - in this case it probably makes sense to enable `feature gps` by default as well so that the GPS will be working out of the box.

mikeller (changes requested):

configs/default/GEPR-GEPRCF722BT.config
arbou: This is a customed new AIO and we tested well,  hope to be approved by you soon.
Merry Christmas!

mikeller: Can you please elaborate a bit on why you are proposing to ship this AIO with integrated GPS that is fully configured but not enabled by default - this seems to be prone to users complaining that their GPS is not working.

mikeller (changes requested):

configs/default/GEPR-GEPRCF722BT.config
arbou: Hi Mikeller, Yes, I missed to add feature gps as default, I will insert a function based on my config, thanks.

configs/default/GEPR-GEPRCF722BT.config
```diff
@@ -0,0 +1,99 @@
...
+dma pin A09 0
+# pin A09: DMA2 Stream 6 Channel 0
+dma pin B03 0
+# pin B03: DMA1 Stream 6 Channel 3
+
+# feature
+feature -RX_PARALLEL_PWM
+feature RX_SERIAL
+feature OSD
+
```
arbou, line 78: `feature gps `

mikeller (approved):

PR #411, opened 2020-12-25
add JHEF411 unified target

sn0opy:
This is the config for the GHF411 PRO AIO.

http://www.jhemcu.com/e_productshow/?17-JHEMCU-GHF411AIO-F4-OSD-FLIGHT-CONTROLLER-BUILT-IN-30A-BL_S-2-4S-4IN1-ESC-FOR-TOOTHPICK-17.html

They ship it with BF 4.2 pre-flashed with a custom target. Unfortunately, they're using "JHE" for the manufacturer ID. Not sure if this is important in some way as BF actually uses "JHEF".

As @atgfpeyv is the official contact for JHE, you might want to have another look into this.

### DIFF
--- a/configs/default/JHEF-JHEF411.config
+++ b/configs/default/JHEF-JHEF411.config
@@ -1,0 +1,113 @@
+# Betaflight / STM32F411 (S411) 4.2.0 Jun 14 2020 / 03:04:43 (8f2d21460) MSP API: 1.43
+
+board_name JHEF411
+manufacturer_id JHEF
+
+# resources
+resource BEEPER 1 C14
+resource MOTOR 1 A08
+resource MOTOR 2 A09
+resource MOTOR 3 A10
+resource MOTOR 4 B00
+resource MOTOR 5 B04
+resource PPM 1 A02
+resource PWM 1 B01
+resource PWM 2 A03
+resource LED_STRIP 1 A15
+resource SERIAL_TX 1 B06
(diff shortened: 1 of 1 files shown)


configs/default/JHEF-JHEF411.config
```diff
@@ -0,0 +1,235 @@
...
+manufacturer_id JHEF
+
+# resources
+resource BEEPER 1 C14
+resource MOTOR 1 A08
+resource MOTOR 2 A09
+resource MOTOR 3 A10
+resource MOTOR 4 B00
+resource MOTOR 5 B04
+resource MOTOR 6 NONE
```
mikeller, line 13: Please read and apply https://github.com/betaflight/betaflight/blob/master/docs/TargetMaintenance/CreatingAUnifiedTarget.md to this - there should not be any lines that are defaults in here.

mikeller (changes requested):
@atgfpeyv - you seem to be the maintainer for JHEF targets - can you review please?

configs/default/JHEF-JHEF411.config
sn0opy: Good point, fixed!

mikeller (approved):
Looking good to me, thanks for contributing!

Wating for a representative of the manufacturer to weigh in, if one can be found.

mikeller:
Mergeing this due to non-response from a maintainer.

PR #414, opened 2021-01-03
add JHEMCU GHF420AIO unified target

NiklasVoigt:
- adc current meter
- adc battery meter
- cw90 gyro alignment


mikeller:
@atgfpeyv - you seem to be the maintainer for JHEF targets - can you review please?

PR #416, opened 2021-01-07
Update GEPR-GEPRCF722BT.config

arbou:
add Bluetooth settings.


mikeller (approved):

mikeller:
@Linjieqiang: Can you please confirm as the maintainer that this change matches with the hardware configuration of this board?

arbou:
We two are for different projects, now this is in charged by me, I can sure it is matching with hardware

mikeller:
@arbou: Cool, should I update this and set you as the maintainer for GEPR?

PR #418, opened 2021-01-08
Add more features for iFlight_F745_AIO board.

Linjieqiang:

PR #422, opened 2021-01-15
Add more features for iFlight_F745_AIO board.

Linjieqiang:
Reopen it. https://github.com/betaflight/unified-targets/pull/418

### DIFF
--- a/configs/default/IFRC-IFLIGHT_F745_AIO.config
+++ b/configs/default/IFRC-IFLIGHT_F745_AIO.config
@@ -9,6 +9,8 @@ resource MOTOR 1 B00
 resource MOTOR 2 B01
 resource MOTOR 3 E09
 resource MOTOR 4 E11
+resource MOTOR 5 B04
+resource MOTOR 6 B05
 resource LED_STRIP 1 D12
 resource SERIAL_TX 1 A09
 resource SERIAL_TX 2 A02
@@ -62,6 +64,10 @@ timer E09 AF1
 # pin E09: TIM1 CH1 (AF1)
 timer E11 AF1
 # pin E11: TIM1 CH2 (AF1)
+timer B04 AF2
+# pin B04: TIM3 CH1 (AF2)
+timer B05 AF2
+# pin B05: TIM3 CH2 (AF2)
 timer D12 AF2
 # pin D12: TIM4 CH1 (AF2)
 timer B10 AF1
@@ -92,6 +98,10 @@ dma pin E09 2
 # pin E09: DMA2 Stream 3 Channel 6
 dma pin E11 1
 # pin E11: DMA2 Stream 2 Channel 6
+dma pin B04 0
+# pin B04: DMA1 Stream 4 Channel 5
+dma pin B05 0
+# pin B05: DMA1 Stream 5 Channel 5
 dma pin D12 0
 # pin D12: DMA1 Stream 0 Channel 2
 dma pin B10 0
@@ -108,11 +118,16 @@ dma pin A03 0
 # feature
 feature -RX_PARALLEL_PWM
 feature RX_SERIAL
+feature TELEMETRY
 feature LED_STRIP
 feature OSD
+feature ESC_SENSOR
 
 # serial
-serial 2 64 115200 57600 0 115200
+serial 0 8192 115200 57600 0 115200
+serial 1 64 115200 57600 0 115200
+serial 5 1024 115200 57600 0 115200
+serial 6 2 115200 57600 0 115200
 
 # led
 led 0 6,11::CTOBVIW:0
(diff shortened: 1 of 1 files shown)


configs/default/IFRC-IFLIGHT_F745_AIO.config
```diff
@@ -107,27 +118,42 @@ dma pin A03 0
 
 # feature
 feature -RX_PARALLEL_PWM
+feature -AIRMODE
```
mikeller, line 121: Why is airmode disabled here? This is a tuning setting, and not part of the hardware configuration.

configs/default/IFRC-IFLIGHT_F745_AIO.config
```diff
@@ -107,27 +118,42 @@ dma pin A03 0
...
 led 3 9,11::CTOBVIW:0
 
+# aux
+aux 0 0 0 900 1300 0 0
+
 # master
 set baro_bustype = I2C
 set baro_i2c_device = 1
-set baro_hardware = BMP280
+set baro_hardware = DPS310
```
mikeller, line 146: Not sure why you are setting this here - leaving this on the default (`AUTO`) will get the device properly.

configs/default/IFRC-IFLIGHT_F745_AIO.config
```diff
@@ -107,27 +118,42 @@ dma pin A03 0
...
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set mag_hardware = HMC5883
```
mikeller, line 149: Not sure why you are setting this here - leaving this on the default (`AUTO`) will get the device properly.

configs/default/IFRC-IFLIGHT_F745_AIO.config
```diff
@@ -1,3 +1,4 @@
+
```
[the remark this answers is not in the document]
Linjieqiang: Fixed it. Thanks.

configs/default/IFRC-IFLIGHT_F745_AIO.config
Linjieqiang: Fixed it. Thanks.

configs/default/IFRC-IFLIGHT_F745_AIO.config
Linjieqiang: Fixed it. Thanks.

configs/default/IFRC-IFLIGHT_F745_AIO.config
Linjieqiang: Fixed it. Thanks.

configs/default/IFRC-IFLIGHT_F745_AIO.config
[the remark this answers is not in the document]
Linjieqiang: Fixed it. Thanks.

configs/default/IFRC-IFLIGHT_F745_AIO.config
```diff
@@ -108,26 +118,40 @@ dma pin A03 0
...
+serial 6 2 115200 57600 0 115200
 
 # led
 led 0 6,11::CTOBVIW:0
 led 1 7,11::CTOBVIW:0
 led 2 8,11::CTOBVIW:0
 led 3 9,11::CTOBVIW:0
 
+# aux
+aux 0 0 0 900 1300 0 0
```
etracer65, line 139: Seems like a very bad idea to pre-configure an arming switch. You have no way of knowing the user's radio setup and aux1 might be used for something else causing an unsafe situation.

configs/default/IFRC-IFLIGHT_F745_AIO.config
```diff
@@ -108,26 +118,40 @@ dma pin A03 0
...
+
 # master
 set baro_bustype = I2C
 set baro_i2c_device = 1
-set baro_hardware = BMP280
+set baro_hardware = AUTO
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set mag_hardware = AUTO
+set align_mag = CW180
```
etracer65, line 148: Does this board actually have an on-board magnetometer? If not then there's no value in setting the alignment as you have no idea how the user will install the external device.

configs/default/IFRC-IFLIGHT_F745_AIO.config
Linjieqiang: Thanks. Removed it now.

configs/default/IFRC-IFLIGHT_F745_AIO.config
Linjieqiang: This board actually have an on-board magnetometer. But I will remove align_mag flag.

etracer65: I suggested removing it if the board **does not** have an onboard sensor. If the board actually has an onboard magnetometer the **you should include the alignment** that's appropriate for sensor's placement on the board.

configs/default/IFRC-IFLIGHT_F745_AIO.config
```diff
@@ -123,11 +138,15 @@ led 3 9,11::CTOBVIW:0
 # master
 set baro_bustype = I2C
 set baro_i2c_device = 1
-set baro_hardware = BMP280
+set baro_hardware = AUTO
```
mikeller, line 141: Please remove the default settings from the configuration.

mikeller (changes requested):

configs/default/IFRC-IFLIGHT_F745_AIO.config
Linjieqiang: Fixed.

mikeller (approved):

PR #423, opened 2021-01-16
Add FPVCYCLEF4 target

TUNERC-Aria:
Add unified target

### DIFF
--- a/configs/default/TURC-FPVCYCLEF401.config
+++ b/configs/default/TURC-FPVCYCLEF401.config
@@ -1,0 +1,73 @@
+# Betaflight / STM32F411 (S411) 4.2.3 Sep 20 2020 / 20:43:12 (2696b7c88) MSP API: 1.43
+
+board_name FPVCYCLEF401
+manufacturer_id TURC
+
+# resources
+resource BEEPER 1 B02
+resource MOTOR 1 B04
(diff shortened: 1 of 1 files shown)


configs/default/TURC-FPVCYCLEF401.config
```diff
@@ -0,0 +1,73 @@
...
+dma pin A02 0
+# pin A02: DMA1 Stream 0 Channel 6
+
+# feature
+feature -RX_PARALLEL_PWM
+feature RX_SERIAL
+feature OSD
+
+# master
+set motor_pwm_protocol = DSHOT600
```
mikeller, line 64: Will this target only be used for an all-in-one offering that comes with on-board ESCs that can handle Dshot600?
If not then using this as a default will lead to the risk of unintended motor spin up if the flight controller is used with certain ESCs.

Furthermore, if this is left at the default, the design of Betaflight configurator is such that the user is prompted to select an appropriate ESC protocol on connect, making this a good user experience, and not likely to result in support requests:

image

TUNERC-Aria: Yes, this FC comes with on-board ESCs that can handle Dshot600.

mikeller (approved):

PR #430, opened 2021-01-25
Added FLCO-FLYCOLORF7_AIO targets

tobeornottobe81:
Added FLCO_FLYCOLORF7_AIO targets


configs/default/FLCO-FLYCOLORF7_AIO.config
```diff
@@ -0,0 +1,144 @@
...
+set baro_i2c_device = 1
+set adc_device = 3
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set current_meter = ADC
+set battery_meter = ADC
+set ibata_scale = 170
+set beeper_inversion = ON
+set beeper_od = OFF
+set pid_process_denom = 4
```
etracer65, line 130: This is certainly wrong. There would be no reason to run the PID loop at 2khz on a F7. Remove this line.

configs/default/FLCO-FLYCOLORF7_AIO.config
```diff
@@ -0,0 +1,144 @@
...
+set dashboard_i2c_bus = 1
+set pinio_box = 40,41,255,255
+set flash_spi_bus = 3
+set gyro_1_bustype = SPI
+set gyro_1_spibus = 1
+set gyro_1_sensor_align = CW180FLIP
+set gyro_1_align_pitch = 1800
+set gyro_1_align_yaw = 1800
+set align_board_yaw = 180
+set align_board_pitch = 180
```
etracer65, line 141: Setting board alignments in the target config doesn't make any sense. It's as if you're telling the user to install the board incorrectly and not match the indicators on the board or documentation. The board alignments should not be set and instead the gyro sensor alignments should be corrected based on the expected board installation orientation.

configs/default/FLCO-FLYCOLORF7_AIO.config
tobeornottobe81: thank you! i will correct it

configs/default/FLCO-FLYCOLORF7_AIO.config
tobeornottobe81: OK, thank you for your guidance

etracer65:
So should we close this pull request? You have removed all the content.

PR #434, opened 2021-01-31
Add StrikerF7

x4FF3:
Hi,

pls add the unified-target for the StrikerF7 board. It will be used for "Soccer Copter".

KR
David

### DIFF
--- a/configs/default/AIRB-HELSEL_STRIKERF7.config
+++ b/configs/default/AIRB-HELSEL_STRIKERF7.config
@@ -1,0 +1,92 @@
+# Betaflight / STM32F7X2 (S7X2) 4.2.2 Aug 16 2020 / 01:48:06 (e833ac612) MSP API: 1.43
+
+board_name HELSEL_STRIKERF7
+manufacturer_id AIRB
+
+# name: StrikerF7
+
+# resources
+resource BEEPER 1 B02
+resource MOTOR 1 C08
+resource MOTOR 2 C09
+resource MOTOR 3 B00
+resource MOTOR 4 B01
+resource LED_STRIP 1 A15
+resource SERIAL_TX 1 A09
+resource SERIAL_TX 2 A02
+resource SERIAL_TX 3 C10
+resource SERIAL_TX 4 A00
+resource SERIAL_TX 5 C12
+resource SERIAL_TX 6 C06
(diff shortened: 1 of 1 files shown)


configs/default/AIRB-HELSEL_STRIKERF7.config
```diff
@@ -0,0 +1,93 @@
...
+
+# master
+set baro_bustype = I2C
+set baro_i2c_device = 1
+set current_meter = ADC
+set battery_meter = ADC
+set ibata_scale = 179
+set beeper_inversion = ON
+set beeper_od = OFF
+set pid_process_denom = 2
```
etracer65, line 87: There is no reason to set the PID loop to 4K on a F722 board. Just remove this and let it default to the correct setting.

configs/default/AIRB-HELSEL_STRIKERF7.config
```diff
@@ -0,0 +1,93 @@
...
+dma pin C09 0
+# pin C09: DMA2 Stream 7 Channel 7
+dma pin B00 0
+# pin B00: DMA1 Stream 7 Channel 5
+dma pin B01 0
+# pin B01: DMA1 Stream 2 Channel 5
+
+# feature
+feature -RX_PARALLEL_PWM
+feature -AIRMODE
```
etracer65, line 73: Please don't disable important flight control features in the target config. This should be up to the user.

mikeller (dismissed):
The target as it is is conforming with the rules for Unified Target configurations - but if you want to incorporate the recommendations from @etracer65 feel free to do so.

configs/default/AIRB-HELSEL_STRIKERF7.config
x4FF3: It is a stripped version of the customers preset for the soccer copter. As the board will be available as spare part, it is a compromise of the full preset which is configured on the soccer copter by default (filter/PIDs). If the sales show, that the board is bought more than expected, we can change this. But atm we would prefer to leave this like it is.

configs/default/AIRB-HELSEL_STRIKERF7.config
x4FF3: I'll remove it.

mikeller (approved):

PR #445, opened 2021-03-07
Update DIAT_FURYF4OSD to enable I2C baro/mag

### LINKED ISSUE #398: External barometer not visible
hello, just bought an external barometer [here](https://www.amazon.fr/HiLetgo-BMP280-3-3-Precision-Atmospheric-Navigation/dp/B07ZWV68M6/ref=sr_1_6?__mk_fr_FR=ÅMÅŽÕÑ&dchild=1&keywords=BMP280&qid=1606070229&sr=8-6), and betaflight do not show on the top right section (it's activate in the configuration menu.
So it's a BME280

i have an F4 Mamba so the target is FURYF4OSD
I solder the SDA to the SDA and the SCL to the SCL
On betaflight 4.2.4
No I2C errors. on the status command.

Do you have any idea why

If you need any more information just tell me.

Thanks

etracer65:
Fixes #398 (and many other github/slack/Facebook issues)

Board has I2C pads but target was not configured to allow users to connect baro/mag resulting in many support issues.

Also removed some other inappropriate settings from the target config:

`set rssi_src_frame_errors = ON` - This is a user preference and also based on the type of radio link used and shouldn't be in the target config.

`set dshot_idle_value = 300` - This is a user preference setting based on the motors/ESCs used. This is a quite low value and could cause motor de-syncs for some users. Board is not an "AIO" sold with motors/ESCs so not appropriate to define this in the target config.

`set pid_process_denom = 1` - This is the default value so it shouldn't be also defined in the target config.

`set gyro_2_spibus = 1` - The board doesn't have a second gyro or even the resources defined for gyro 2.

### DIFF
--- a/configs/default/DIAT-FURYF4OSD.config
+++ b/configs/default/DIAT-FURYF4OSD.config
@@ -83,16 +83,17 @@ feature OSD
 serial 2 8192 115200 57600 0 115200
 
 # master
-set rssi_src_frame_errors = ON
 set serialrx_provider = SBUS
 set blackbox_device = SPIFLASH
-set dshot_idle_value = 300
 set motor_pwm_protocol = DSHOT600
 set current_meter = ADC
 set battery_meter = ADC
 set beeper_inversion = ON
 set beeper_od = OFF
-set pid_process_denom = 1
+set baro_bustype = I2C
+set baro_i2c_device = 1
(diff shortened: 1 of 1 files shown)


kc10kevin (approved):
Looks good.   Thanks for the input.

PR #448, opened 2021-03-20
Added Target "IFRC-IFLIGHT_F745_AIO_V2"

Linjieqiang:


configs/default/IFRC-IFLIGHT_F745_AIO_V2.config
```diff
@@ -0,0 +1,136 @@
...
+# feature
+feature -RX_PARALLEL_PWM
+feature -AIRMODE
+feature RX_SERIAL
+feature TELEMETRY
+feature LED_STRIP
+feature OSD
+
+# master
+set align_mag = CW180
```
mikeller, line 115: Does this board come with an on-board magnetometer?

Linjieqiang: Yes, our current boards are already equipped with an onboard magnetic compass as standard.

mikeller (approved):

PR #456, opened 2021-04-06
Delete added motors

kc10kevin:
Michael.....Diatone changed their mind (found errors) and would like to revert back to the original config files.  Sorry for the issues.

Thanks so much!


kc10kevin:
@mikeller Hi Michael....sorry to bother since you are extremely busy, but Diatone is wondering if these changes might be able to be incorporated soon.  Seems there is a conflict.   Thanks in advance for your time!   Kevin

mikeller (approved):

PR #465, opened 2021-05-12
Add target GEPR_GEPRCF411_AIO

Linjieqiang:


configs/default/GEPR_GEPRCF411_AIO.config
```diff
@@ -0,0 +1,119 @@
...
+# pin A02: DMA1 Stream 0 Channel 6
+dma pin A08 0
+# pin A08: DMA2 Stream 6 Channel 0
+
+# feature
+feature -RX_PARALLEL_PWM
+feature RX_SERIAL
+feature SOFTSERIAL
+feature OSD
+feature AIRMODE
```
mikeller, line 88: Please do not explicitly enable default features: `AIRMODE`, `ANTI_GRAVITY`, `DYNAMIC_FILTER`
This will result in the configuration failing in future versions of the firmware.

mikeller (changes requested):

configs/default/GEPR_GEPRCF411_AIO.config
Linjieqiang: Rebased and fixed it. Thanks.

mikeller (approved):

configs/default/GEPR_GEPRCF411_AIO.config
```diff
@@ -0,0 +1,116 @@
+# Betaflight / STM32F411 (S411) 4.1.0 Jun 25 2019 / 10:27:57 (2a6e94d03) MSP API: 1.42
```
mikeller, line 1: Please fix the file name - should be `GEPR-GEPRCF411_AIO.config`.

mikeller (changes requested):

configs/default/GEPR_GEPRCF411_AIO.config
Linjieqiang: fixed it.

haslinghuis (approved):

mikeller (approved):

PR #467, opened 2021-05-20
Add flowbox target

DusKing1:
This was made for RSgoBlackbox community.
Only have:
- a flash chip,
- a beeper,
- an external button,
- a BMI270 gyro

and no motor output.
image
image

### DIFF
--- a/configs/default/NEUT-FLOWBOX.config
+++ b/configs/default/NEUT-FLOWBOX.config
@@ -5,7 +5,6 @@ manufacturer_id NEUT
 
 # resources
 resource BEEPER 1 B01
-resource LED_STRIP 1 A15
 resource SERIAL_TX 1 A09
 resource SERIAL_TX 2 A02
 resource SERIAL_RX 1 A10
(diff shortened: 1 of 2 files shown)


configs/default/NEUT-FLOWBOX.config
```diff
@@ -0,0 +1,64 @@
...
+resource SPI_MOSI 3 B05
+resource FLASH_CS 1 A15
+resource GYRO_EXTI 1 A01
+resource GYRO_CS 1 A04
+resource PWM 1 B10
+
+# timer
+timer B01 AF2
+# pin B01: TIM3 CH4 (AF2)
+timer A15 AF1
```
mikeller, line 28: What's the timers / DMAs used for in addition to the beeper?
Also, the PWM pin will need a timer to work.

mikeller:
Interesting idea. :grinning: 

Happy to support this for now, but obviously this is outside of the range of hardware that Betaflight is targeting. So if future changes to Betaflight mean that this won't work with a future version of Betaflight there probably won't be any attempts to make this work again. But then I guess since this isn't really dependent on flight performance improvements / new flight related features, there is no need for this hardware to always run the latest version of Betaflight.

DusKing1:
Yes, this make sense, but only a few features was involved in this hardware, it's quite hard to say in the future it will be outdated.

configs/default/NEUT-FLOWBOX.config
DusKing1: For serial receivers. Thanks for your hint, will update.

rhys100:
Hey @DusKing1 where does one buy these chips?

DusKing1:
@rhys100 currently this hardware didn't get massive manufactured due to the chip lack. You can stay tuned to the RSgoBlackbox community in discord, or AliExpress.

PR #468, opened 2021-05-22
Add target CUST RUSRACE_F4 and RUSRACE_F7

winkingsl:


configs/default/CUST-RUSRACE_F4.config
```diff
@@ -0,0 +1,114 @@
+# Betaflight / STM32F405 (S405)
+
+board_name RUSRACE_F4
```
mikeller, line 3: Is there any documentation for this target available?

configs/default/CUST-RUSRACE_F4.config
```diff
@@ -0,0 +1,114 @@
...
+feature LED_STRIP
+feature OSD
+
+# serial
+serial 0 64 115200 57600 0 115200
+serial 1 1024 115200 57600 0 115200
+serial 2 8192 115200 57600 0 115200
+
+# master
+set mag_hardware = NONE
```
mikeller, line 100: This is not required - mag and baro autodetect works just fine if these are left as defaults.

configs/default/CUST-RUSRACE_F7.config
```diff
@@ -0,0 +1,113 @@
...
+set mag_hardware = NONE
+Set baro_hardware = NONE
+set serialrx_provider = SBUS
+set blackbox_device = SPIFLASH
+set current_meter = ADC
+set battery_meter = ADC
+set max7456_spi_bus = 2
+set displayport_max7456_wht = 3
+set pinio_config = 129,1,1,1
+set pinio_box = 39,0,0,0
```
mikeller, line 108: What is this pin used for?

configs/default/CUST-RUSRACE_F4.config
winkingsl: https://goo.su/RusRace

configs/default/CUST-RUSRACE_F4.config
winkingsl: there is no i2c pins on board

configs/default/CUST-RUSRACE_F7.config
winkingsl: hardware pit mode

mikeller: This will cause pit mode to not work correctly for every VTX that has support for native pit mode, as this is the same box number as VTX pit mode. Please use a different box id for this. Most boards use 40.

configs/default/CUST-RUSRACE_F4.config
mikeller: Nice, thank you.

configs/default/CUST-RUSRACE_F4.config
mikeller: Even without I2C pins leaving this at the default will work just fine. :wink:

winkingsl: I just read the instructions=) `(Exception: Boards that do not have any pins or pads for an SPI or I2C bus exposed can define these settings as NONE.)`

configs/default/CUST-RUSRACE_F7.config
winkingsl: i was change it

winkingsl:
@mikeller i need to do something else?

mikeller (approved):

PR #471, opened 2021-06-15
FLYWOO F745NANO FC update targets config

flywoo:
FLYWOO F745NANO FC update targets config


configs/default/FLWO-FLYWOOF745NANO.config
```diff
@@ -17,20 +17,22 @@ resource PPM 1 E13
 resource LED_STRIP 1 D12
 resource SERIAL_TX 1 A09
 resource SERIAL_TX 2 D05
-resource SERIAL_TX 3 B10
```
mikeller, line 20: Has a version of this board already been sold with the old configuration?

flywoo: The previously uploaded config is wrong. There is no problem with the FC PCB that has been sold. thank

mikeller (approved):

PR #473, opened 2021-06-20
Add NeutronRC F7 AIO

DusKing1:
AIO board, ESCs are onboard. So we will take response for the Dshot settings

image

### DIFF
--- a/configs/default/NEUT-NEUTRONRCF7AIO.config
+++ b/configs/default/NEUT-NEUTRONRCF7AIO.config
@@ -1,6 +1,6 @@
 # Betaflight / STM32F745 (S745) 4.1.0 Sep 25 2019 / 01:23:16 (064ca9b75) MSP API: 1.42
 
-board_name NEUTRON_F7_AIO
+board_name NEUTRONRCF7AIO
 manufacturer_id NEUT
 
 # resources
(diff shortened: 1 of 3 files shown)


DusKing1:
@github-actions get away

configs/default/NEUT-NEUTRONRCF7AIO.config
```diff
@@ -0,0 +1,147 @@
+# Betaflight / STM32F745 (S745) 4.1.0 Sep 25 2019 / 01:23:16 (064ca9b75) MSP API: 1.42
+
+board_name NEUTRONRCF7AIO
+manufacturer_id NEUT
```
mikeller, line 4: There does not seem to be a manufacturer with this ID. Can you please follow [these instructions](https://github.com/betaflight/betaflight/blob/master/docs/TargetMaintenance/CreatingAUnifiedTarget.md#23-add-the-board-and-manufacturer-information) to get one added?

configs/default/NEUT-NEUTRONRCF7AIO.config
```diff
@@ -0,0 +1,147 @@
...
+feature OSD
+feature ESC_SENSOR
+
+# serial
+serial 0 8192 115200 57600 0 115200
+serial 1 64 115200 57600 0 115200
+serial 5 1024 115200 57600 0 115200
+
+# aux
+aux 0 0 0 900 1300 0 0
```
mikeller, line 120: This should not be in here.

configs/default/NEUT-NEUTRONRCF7AIO.config
DusKing1: OMG, I haven't awared that my ID is no longer the NEUT but NERC

configs/default/NEUT-NEUTRONRCF7AIO.config
DusKing1: Will fix this

mikeller (approved):

PR #474, opened 2021-07-09
Update RAST-AIRF7.config

funfminuten:
Default inverted pinio_config makes the vtx active on startup at least for a split second. Making it non inverted solves the issue - vtx is only active on demand.


mikeller:
This needs a rebase.

mikeller (approved):

PR #466, opened 2021-05-15
Fix ADC stream selection to allow for SPI DMA from gyro

SteveCEvans:
On F7 processors, when bi-directional DSHOT is enabled, depending on motor pin assignment you may see the following in https://github.com/betaflight/betaflight/pull/10705.

```
# dma show

Currently active DMA:
--------------------
DMA1 Stream 0: SDCARD_CS
DMA1 Stream 1: FREE
DMA1 Stream 2: FREE
DMA1 Stream 3: OSD_CS
DMA1 Stream 4: OSD_CS
DMA1 Stream 5: SDCARD_CS
DMA1 Stream 6: FREE
DMA1 Stream 7: FREE
DMA2 Stream 0: ADC
DMA2 Stream 1: FREE
DMA2 Stream 2: DSHOT_BITBANG 2
DMA2 Stream 3: GYRO_CS
DMA2 Stream 4: FREE
DMA2 Stream 5: FREE
DMA2 Stream 6: FREE
DMA2 Stream 7: FREE
```

Note that there is only one stream owned by GYRO_CS, but two are required for each SPI bus, one for transmit and one for receive. This is needed to support https://github.com/betaflight/betaflight/pull/10573 that follows https://github.com/betaflight/betaflight/pull/10705. To fix this issue, we can simply reassign the ADC to use an alternate stream thus.

```
# dma ADC 1 1
# dma ADC 1: changed from 0 to 1
```

Following a reboot you'll then see.

```
# dma show

Currently active DMA:
--------------------
DMA1 Stream 0: SDCARD_CS
DMA1 Stream 1: FREE
DMA1 Stream 2: FREE
DMA1 Stream 3: OSD_CS
DMA1 Stream 4: OSD_CS
DMA1 Stream 5: SDCARD_CS
DMA1 Stream 6: FREE
DMA1 Stream 7: FREE
DMA2 Stream 0: GYRO_CS
DMA2 Stream 1: FREE
DMA2 Stream 2: DSHOT_BITBANG 2
DMA2 Stream 3: GYRO_CS
DMA2 Stream 4: ADC
DMA2 Stream 5: FREE
DMA2 Stream 6: FREE
DMA2 Stream 7: FREE
```

This change is known to be needed for the MATEKF722SE and SPEEDYBEEF7V2.

PR #478, opened 2021-07-26
Revert DIAT 466

kc10kevin:
Diatone contacted me for emergency change to these.   Boards becoming inoperative due to #466 .

Revert All Diatone 466 changes.  Boards become inoperative once config uploaded.   I am moving right now and everything packed up, but all I can surmise is that the change conflicts with TIM8_UP on Stream 1, locking up the board.  Please change this back ASAP so we can get the boards working again for folks and then figure out what is going on.  Sorry about this.  Should have noticed earlier.

### DIFF
--- a/configs/default/DIAT-MAMBAF405US.config
+++ b/configs/default/DIAT-MAMBAF405US.config
@@ -90,8 +90,8 @@ resource LED_STRIP 1 B03
 resource CAMERA_CONTROL 1 B08
 
 # DMA
-dma ADC 3 1
-# ADC 3: DMA2 Stream 1 Channel 2
+dma ADC 3 0
+# ADC 3: DMA2 Stream 0 Channel 2
 dma pin A09 0
 # pin A09: DMA2 Stream 6 Channel 0
 dma pin A08 0
--- a/configs/default/DIAT-MAMBAF405US_I2C.config
+++ b/configs/default/DIAT-MAMBAF405US_I2C.config
@@ -91,8 +91,8 @@ resource MOTOR 4 C08
 resource LED_STRIP 1 B03
 
 # DMA
-dma ADC 3 1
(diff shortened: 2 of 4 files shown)


mikeller (approved):

mikeller:
@kc10kevin: Fixed.

@SteveCEvans: Do you have an idea what is going on here? I think it will be helpful to get to the bottom of this, in order to see if there are other targets that have the same problem with the ADC config changes. I have got these boards and I can test with them.

SteveCEvans:
Any specific FC?

mikeller:
Ok, I figured out why this resulted in a wedge:

```
# dma show

Currently active DMA:
--------------------
DMA1 Stream 0: FREE
DMA1 Stream 1: FREE
DMA1 Stream 2: FREE
DMA1 Stream 3: FREE
DMA1 Stream 4: FREE
DMA1 Stream 5: FREE
DMA1 Stream 6: FREE
DMA1 Stream 7: FREE
DMA2 Stream 0: ADC
DMA2 Stream 1: TIMUP 8   <= This is where ADC 3 was going after the reconfiguration
DMA2 Stream 2: FREE
DMA2 Stream 3: FREE
DMA2 Stream 4: FREE
DMA2 Stream 5: TIMUP 1
DMA2 Stream 6: FREE
DMA2 Stream 7: FREE
```

Unfortunately the firmware does not detect DMA conflicts and in case of conflict does a double allocation, resulting in a wedge of the firmware in most such cases.
We need to add code to detect and avoid conflicts.

Luckily it looks like there might be no more targets that have such a conflict:

```
mikeller@pc unified-targets % grep -l "DMA2 Stream 1" configs/default/* | sort | uniq -c | grep -v "^   1 "
```

PR #490, opened 2021-08-08
Added check for the correct file encoding.

mikeller:
This is to reject files that are using DOS line endings.

Also moved the shell script based checks into a shell script for easier maintenance.


haslinghuis (approved):
Nice!

PR #492, opened 2021-08-09
Set the target BETAFPVF411RX dshot_bitbang to OFF

betafpv-engineer:
This modification is to synchronize [betaflight/unified-targets#95](https://github.com/betaflight/unified-targets/pull/95). 

Solve the conflict between use timer dshot and timer1.


mikeller (approved):
Thanks for addressing this.

PR #496, opened 2021-08-19
Added SOFTSERIAL2 to the BETAFPVF411 target

betafpv-engineer:
PB10->SOFTSERIAL2.


haslinghuis (approved):

Asizon:
This is wrong, there are RX and TX using same resource.

haslinghuis:
Original pull is 133.

mikeller:
@Asizon: This is a supported configuration - obviously it will not work for full duplex connections, but peripherals that use half duplex communication on a single pin will work.

PR #499, opened 2021-08-21
Add target for IFRC/IFLIGHT_BLITZ_F405

XING-IF:
This is a first time submission of a new target config, please bare with me.
Would be nice if there's no further need for a derailing discussion or chinese social media posts (Thanks DusKing!)
Productive feedback always wanted, please check if correct!

To the instruction template:
Point2 I saw that many commitments use the master branch, and then point3 compare to the master as well. Is that correct?

Point5: I've uploaded the new file, forgot to add the name "IFRC-" so I've reuploaded the file. Is that ok?


DusKing1:
Why you can't stop spam? I have talked to your boss and she don't want you continuely submit wrong target.

XING-IF:
I've uploaded #500. The #499 had a file issue when uploading obviously.
Be more productive, you're really hard to work with and unable to talk to
professionally seriously...

PR #501, opened 2021-08-23
Added BLITZ F405 Target

Si-iF:
Initial commit:
Added F405 target for Iflight BLIZT F405

@mikeller This is iFlight new official Github account for target. 
We would like to be code maintainers for our target from now on. We will also maintain our old target. Please kindly modify the code owner for IFRC if possible.
Should we open an issue for modify the code ownership of IFRC target under this account or can you do it without any action from us?

### DIFF
--- a/configs/default/IFRC-IFLIGHT_BLITZ_F405.config
+++ b/configs/default/IFRC-IFLIGHT_BLITZ_F405.config
@@ -1,0 +1,97 @@
+# Betaflight / STM32F405 (S405) 4.2.9 Apr 27 2021 / 19:33:01 (e097f4ab7) MSP API: 1.43
+
+board_name IFLIGHT_BLITZ_F405 
+manufacturer_id IFRC
+
+# resources
+resource BEEPER 1 B02
+resource MOTOR 1 B00
+resource MOTOR 2 B01
+resource MOTOR 3 C09
+resource MOTOR 4 C08
+resource PPM 1 A03
+resource LED_STRIP 1 A08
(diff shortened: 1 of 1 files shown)


configs/default/IFRC-IFLIGHT_BLITZ_F405.config
```diff
@@ -0,0 +1,97 @@
+# Betaflight / STM32F405 (S405) 4.2.9 Apr 27 2021 / 19:33:01 (e097f4ab7) MSP API: 1.43
```
DusKing1, line 1: This is still exported from third-party firmware, not from our official unified target firmware. How many times do you wanna waste on us? How many times do I need to teach you?

DusKing1 (changes requested):
This is the third time of your guys' "first time", and the issue still exists? When can you stop wasting everybody's time?

configs/default/IFRC-IFLIGHT_BLITZ_F405.config
Guidus93: After a call with @DusKing1 we found out that everything is fine on the version. The config file is exported from official STM32F405 4.2.9 from April 27 19.33 . Looks all good here finally

WeChat Image_20210823173025

Guidus93:
@haslinghuis @Linjieqiang i check the config file and tested it on the prototype. Everything looks ok. Can you guys approve this PR?

haslinghuis (approved):

PR #503, opened 2021-08-28
Add EACHINEF411_AIO.config

EachineRC:
Add EACHINEF411_AIO.config

### DIFF
--- a/configs/default/EACH-EACHINEF411_AIO.config
+++ b/configs/default/EACH-EACHINEF411_AIO.config
@@ -3,8 +3,6 @@
 board_name EACHINEF411_AIO
 manufacturer_id EACH
 
-# name
-
 # resources
 resource BEEPER 1 B02
 resource MOTOR 1 B05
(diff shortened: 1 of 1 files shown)


configs/default/EACH-EACHINEF411_AIO.config
```diff
@@ -0,0 +1,92 @@
+# Betaflight / STM32F411 (S411) 4.2.9 Apr 27 2021 / 19:33:23 (e097f4ab7) MSP API: 1.43
+
+board_name EACHINEF411_AIO
+manufacturer_id EACH
+
+# name
```
DusKing1, line 6: Should be removed

DusKing1 (changes requested):

configs/default/EACH-EACHINEF411_AIO.config
EachineRC: Thanks for your reminder.

DusKing1 (approved):

mikeller (approved):

PR #514, opened 2021-10-04
Added Unified Target for IFLIGHT_H7_TWING.

mikeller:


DusKing1:
It seems that the flash chips on QUAD-SPI can not work on H743 unified target.

DusKing1 (approved):
It seems that the flash chips on QUAD-SPI can not work on H743 unified target. Other than that all seems to work properly.

mikeller:
@DusKing1: Yes, unfortunately quad SPI is poor at the moment and does not properly support Unified Targets. Does the IFlight H7 TwinG have a quad SPI attached flash chip?

DusKing1:
@mikeller AFAIK yes, it's the W25N01G, and if you have the same sample as me, on the bottom side there are only two chips, one is MAX7456 and the other one is that flash chip.

PR #521, opened 2021-10-25
Fix azure pipelines

DusKing1:
https://docs.microsoft.com/en-us/azure/devops/release-notes/2021/pipelines/sprint-193-update#updated-schedule-for-removal-of-ubuntu-1604-image-on-microsoft-hosted-agents

@daleckystepan Give this a review?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,6 +1,9 @@
 # Runs the checks for Unified Target configurations
 #
 
+variables:
+  vmImage: "ubuntu-20.04"
+
 name: $(Build.RequestedFor)_$(Build.SourceBranchName)_$(BuildID)
 
 trigger:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -1,6 +1,11 @@
 # Runs the checks for Unified Target configurations
 #
 
+variables:
+  owner: betaflight
```
mikeller, line 5: What are `owner` and `repoName` used for?

DusKing1: IDK, I was just following other repos' yml style. Shall we remove this?

mikeller (approved):

azure-pipelines.yml
mikeller: Yes please, thanks. Keeping unused variables around makes the code harder to understand and maintain.

PR #522, opened 2021-10-29
Create config file for BLITZ F722

Si-iF:
Initial commit:
Added config file for IFRC-BLITZ_F722 FC


mikeller (approved):
Ran the check scripts manually - all ok according to them.

Issue #523, opened 2021-10-31
HGLRC Zeus 722 hangs after applying config.


obeny:
Description:
HGLRC Zeus F722 hangs after power on.

How to reproduce:
Flash latest 4.3 development BF, apply custom defaults when asked.
Board is non-responsive, switching tabs in configurator is not possible.

This problem affects only Zeus F722 FCs only.

For details see:
https://github.com/betaflight/betaflight/issues/11026

HGLRC-D:
Do not! This will make the BMP280 on the target not work properly.

jessearmand:
Hi @obeny and @HGLRC-D this default configuration is not right, I'm using firmware version 4.2.10, and I had to change the `baro_bustype` into SPI. That's how it was configured before this unified target changes.

Here's what I set to enable the barometer again:
````
set baro_bustype = SPI
set baro_spi_device = 1
set baro_i2c_device = 0
````

jessearmand:
@HGLRC-D As I mentioned elsewhere, I noticed that this config was created without `baro_bustype` being set to any value. 

As can be seen here in [barometer.c](https://github.com/betaflight/betaflight/blob/18af30321dc6ba352f44125fd923f359560a8e5a/src/main/sensors/barometer.c#L155-L273) not having a value there could potentially cause problems. 

But, I haven't tested on the latest 4.3, could you please verify that the BMP280 should work normally with SPI?

alienwaredream:
> Do not! This will make the BMP280 on the target not work properly.

所以在BF4.3下化骨龙F722飞控到底能不能启用气压计呢？

HGLRC-D:
> > 不要！这将使目标上的 BMP280 无法正常工作。
> 
> 所以在BF4.3下化骨龙F722飞控到底能不能启用气压计呢？

可以用的

DusKing1:
English please?

alienwaredream:
> > > 不要！这将使目标上的 BMP280 无法正常工作。
> > 
> > 
> > 所以在BF4.3下化骨龙F722飞控到底能不能启用气压计呢？
> 
> 可以用的

我试了在4.2下用

HGLRC-D:
I tested 4.3 here and there is no problem. Have you tried to re-burn the firmware?

> English please?

Sorry, i will pay attention.

HGLRC-D:
> > > > 不要！这将使目标上的 BMP280 无法正常工作。
> > > 
> > > 
[...]

PR #524, opened 2021-10-31
HGLRC Zeus F722 - fix baro config

obeny:
Baro BMP280 is connected to the I2C bus, not SPI.
This fixes FC hangup (#523).

### DIFF
--- a/configs/default/HGLR-HGLRCF722.config
+++ b/configs/default/HGLR-HGLRCF722.config
@@ -113,7 +113,8 @@ serial 0 64 115200 57600 0 115200
 # master
 set mag_bustype = I2C
 set mag_i2c_device = 1
-set baro_spi_device = 1
+set baro_bustype = I2C
+set baro_i2c_device = 1
 set blackbox_device = SPIFLASH
 set dshot_burst = ON
 set motor_pwm_protocol = DSHOT600


DusKing1 (approved):

jessearmand:
@mikeller @obeny This default configuration is not right. I'm using firmware version 4.2.10, and I had to change the `baro_bustype` into SPI. That's how it was configured before this unified target changes.

Here's what I set to enable the barometer again:

```
set baro_bustype = SPI
set baro_spi_device = 1
set baro_i2c_device = 0
```

@HGLRC-D please clarify whether it's an SPI or I2C bus type?

obeny:
Honestly, haven't verified it against 4.2.x. Stopped using it like 8 months ago and almost forgot about it's existence :)
I'm pretty sure I saw baro detected. However I'm not using it on any of my quads.
On 4.3 this change fixes FC hang on powerup. I'll take a deeper look at it on Sunday. Without this change 4.3 is unusable on Zeus F722. For now I think it is better to have FC booting, while having problem with baro than completely unusable FC.

jessearmand:
You mentioned that this FC hang up is because of the recent change to
support dual gyro on a shared SPI bus.

Without that change, I believe it should work fine for baro with SPI.

Also does having no `baro_bustype` being set to any value (as such is the
defaults before you made this change to i2C) is normal? But it wasn't like
that originally.

I  will test this on 4.3

jessearmand:
I have just tested on 4.3, and it does hang up if I set it this way. If I revert back to https://github.com/betaflight/betaflight/commit/192799582c248962ab4f314edce6faaecdfab332 then baro is working with SPI. 

But, now I see issues with IRC Ghost after upgrading, stick commands are not responding, although it's bound to my Ghost TX, switched to SRXL2 same issues. Now, I'm troubleshooting another issue.

So, the baro in my FC does indeed use SPI, but probably I should stick with 4.2.x 😢 

@HGLRC-D Any comments?

SteveCEvans:
Just checked the config for this device...

```
set baro_spi_device = 1
```

And...
```
set max7456_spi_bus = 2
set flash_spi_bus = 3
set gyro_1_bustype = SPI
set gyro_1_spibus = 1
```

Bad design choice there. The gyro really wants to be on a bus by itself. Nevertheless the SPI bus driver shouldn't hang. I'll check my FC collection to see if I have anything similarly configured and debug.

obeny:
Steve if you haven't got anything similar, I can do some debugging.

Just played with the settings and baro is not detected. I was wrong, BMP280 is on SPI bus. Sorry for the confusion. I didn't care about baro, not using it.

PR #527, opened 2021-11-12
Add new target NEUTRONRCF407

DusKing1:
The chip is STM32F407, we have verified that this can work with F405 unified target. But the only thing not sure is whether we need to not use this "F407" name that might cause potential misunderstand

### DIFF
--- a/configs/default/NERC-NEUTRONRCF407.config
+++ b/configs/default/NERC-NEUTRONRCF407.config
@@ -1,0 +1,120 @@
+# Betaflight / STM32F405 (S405) 4.0.0 Mar 14 2019 / 11:45:26 (360afd96d) MSP API: 1.41
+
+board_name NEUTRONRCF407
+manufacturer_id NERC
+
+# resources
+resource BEEPER 1 B04
+resource MOTOR 1 B00
+resource MOTOR 2 B01
+resource MOTOR 3 C09
+resource MOTOR 4 C08
+resource PPM 1 A03
+resource LED_STRIP 1 B06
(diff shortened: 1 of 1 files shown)


configs/default/NERC-NEUTRONRCF407.config
```diff
@@ -0,0 +1,120 @@
...
+set battery_meter = ADC
+set ibata_scale = 100
+set beeper_inversion = ON
+set beeper_od = OFF
+set max7456_spi_bus = 2
+set dashboard_i2c_bus = 1
+set flash_spi_bus = 3
+set gyro_1_bustype = SPI
+set gyro_1_spibus = 1
+set gyro_1_sensor_align = CW270
```
limonspb, line 120: new line

configs/default/NERC-NEUTRONRCF407.config
```diff
@@ -0,0 +1,120 @@
...
+feature TELEMETRY
+feature LED_STRIP
+feature OSD
+
+# serial
+serial 0 8192 115200 57600 0 115200
+serial 1 64 115200 57600 0 115200
+
+# master
+set serialrx_provider = SBUS
```
limonspb, line 102: is that AIO?

limonspb (changes requested):

configs/default/NERC-NEUTRONRCF407.config
DusKing1: no

haslinghuis (approved):

PR #533, opened 2021-12-02
Add IFRC F411 RX target

Si-iF:
Add IFlight F411 RX (spi rx) target.


configs/default/IFRC-IFLIGHT_BLITZ_F411RX.config
```diff
@@ -0,0 +1,92 @@
...
+set rx_spi_bus = 3
+set rx_spi_led_inversion = ON
+set dshot_bitbang = OFF
+set current_meter = ADC
+set battery_meter = ADC
+set ibata_scale = 180
+set system_hse_mhz = 8
+set max7456_spi_bus = 2
+set cc2500_spi_chip_detect = OFF
+set dashboard_i2c_bus = 1
```
DusKing1, line 88: This is not supported, should be removed

DusKing1 (changes requested):

DusKing1:
https://github.com/betaflight/betaflight/blob/master/CODE_OF_CONDUCT.md

PR #534, opened 2021-12-03
Update GEPR-GEPRCF722BT.config for fix voltage meter display error

flosean:
Help GEPRC and (VGOOD made this board) to fix battery  voltage meter display error.
In 4.1 battery voltage meter display correctly, but in 4.2 it broken. They find change the ADC DMA number can fix it.

### DIFF
--- a/configs/default/GEPR-GEPRCF722BT.config
+++ b/configs/default/GEPR-GEPRCF722BT.config
@@ -1,4 +1,4 @@
-# Betaflight / STM32F7X2 (S7X2) 4.2.11 Nov  9 2021 / 20:29:32 (948ba6339) MSP API: 1.43
+# Betaflight / STM32F7X2 (S7X2) 4.1.7 May 28 2020 / 15:06:20 (9ba02a587) MSP API: 1.42
 board_name GEPRCF722BT
 manufacturer_id GEPR
 
--- a/configs/default/GEPR-GEPRCF722BT.config
+++ b/configs/default/GEPR-GEPRCF722BT.config
@@ -1,4 +1,4 @@
-# Betaflight / STM32F7X2 (S7X2) 4.1.7 May 28 2020 / 15:06:20 (9ba02a587) MSP API: 1.42
+# Betaflight / STM32F7X2 (S7X2) 4.2.11 Nov  9 2021 / 20:29:32 (948ba6339) MSP API: 1.43
 board_name GEPRCF722BT
 manufacturer_id GEPR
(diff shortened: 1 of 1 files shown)


haslinghuis:
Does this board feature an on-board GPS?

flosean:
> Does this board feature an on-board GPS?

Just check this board doesn't have onboard GPS module, remove default GPS mode turn on.

configs/default/GEPR-GEPRCF722BT.config
```diff
@@ -1,4 +1,4 @@
-# Betaflight / STM32F7X2 (S7X2) 4.1.7 May 28 2020 / 15:06:20 (9ba02a587) MSP API: 1.42
```
DusKing1, line 1: what's the reason for changing API version? This will make users can't use older firmware anymore.

DusKing1 (changes requested):

configs/default/GEPR-GEPRCF722BT.config
flosean: Because they use new firmware to make the unified target and verify it, API version update itself.

DusKing1: This will force every user who uses 4.1.x to upgrade to 4.2.x. Shouldn't be changed.

flosean: Edit it back to 1.42.

haslinghuis (approved):

DusKing1:
merge without squash?

haslinghuis:
Max of 3 commits. Squash would be nice.

configs/default/GEPR-GEPRCF722BT.config
```diff
@@ -58,8 +58,8 @@ timer B03 AF1
 # pin B03: TIM2 CH2 (AF1)
 
 # dma
-dma ADC 3 1
-# ADC 3: DMA2 Stream 1 Channel 2
+dma ADC 3 0
```
mikeller, line 61: @SteveCEvans: Another target that was changed back to a brolen SPI DMA configiuration - do you have time to figure out how this target can be fixed?

@haslinghuis: Please do not merge problematic changes like this.

flosean: I ckeched with them, they said that change the ADC dma work fine on 4.2. But with default define it will broken on 4.2, still doesn't know why.

PR #538, opened 2021-12-09
Fix boot_grace_time flag on flowbox target

DusKing1:
Added motor settings for flowbox target, confirmed to work. We might make flowbow variants with motor pads anyway in the future, so this is needed.


haslinghuis:
It would be better to have all motor resource on same dma channel and timer. Or split 2 - 2

DusKing1:
@haslinghuis can you elaborate yourself about the "better" section?

haslinghuis (approved):

PR #543, opened 2021-12-22
disable GYRO_EXTI when Gyro and OSD are on the same SPI bus

ctzsnooze:
I found that the following boards also share the Gyro and OSD on the same SPI bus, and may be vulnerable to the same issue reported in https://github.com/betaflight/betaflight/pull/11141

> LEGA-KROOZX - both on 1
> YUPF-YUPIF4 - both on 1
> YUPF-YUPIF7 - both on 1
> ZEEZ-ZEEZWHOOP - both on 1
> AIRB-NOX - both on 2
> AIRB-OMNIBUSF4F - both on 3

As a safety precaution, this PR disables GYRO_EXTI on those boards.

I include a comment so the reason why is known.

### DIFF
--- a/configs/default/AIRB-NOX.config
+++ b/configs/default/AIRB-NOX.config
@@ -29,7 +29,9 @@ resource ADC_BATT 1 A05
 resource BARO_CS 1 A09
 resource FLASH_CS 1 A15
 resource OSD_CS 1 A10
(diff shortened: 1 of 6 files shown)


ctzsnooze:
messed up will re-submit

PR #555, opened 2022-01-16
config fix MATEKH743

TheIsotopes:
- removed User Settings
- set feature to defaults

@SteveCEvans this reflects the configuration 100% to the legacy target.

@MATEKSYS can you confirm this.

### DIFF
--- a/configs/default/MTKS-MATEKH743.config
+++ b/configs/default/MTKS-MATEKH743.config
@@ -159,12 +159,11 @@ dma pin B09 0
 # feature
 feature -RX_PARALLEL_PWM
 feature RX_SERIAL
-feature LED_STRIP
+feature TELEMETRY
 feature OSD
 
 # serial
 serial 5 64 115200 57600 0 115200
-serial 7 1024 115200 57600 0 115200
 
 # master
 set mag_bustype = I2C


haslinghuis (approved):

blckmn (approved):

SteveCEvans:
We can turn off LED_STRIP and enable TELEMETRY if that’s what’s desired. Both on night make good sense. It’s not like there will be a CPU load issue.

SteveCEvans:
UART8 is for ESC Serial that should have been left in place. It was an omission on the legacy.

TheIsotopes:
> UART8 is for ESC Serial that should have been left in place. It was an omission on the legacy.

@SteveCEvans I think if someone needs this they can it setup.
imho the unified targets should be comparable to the legacy target and no extras should be installed.

SteveCEvans:
@TheIsotopes  the config should reflect what the user sees in the manual or what’s the point of the config? If the ESC cable is plugged in then this is the function for UART8. As a user, knowing the FC supports ESC_SENSOR I expect to turn it on in the configurator and for it to work.

SteveCEvans:
Also, given that we all know “two wrongs don’t make a right” once we’ve agreed what’s best for the user with the most comprehensive description of the FC in the config, we should raise an issue against the legacy target to fix it; 4.3.1 I suggest.

SteveCEvans:
And having given the LED_STRIP and  TELEMETRY features some more thought; we should follow convention here. If other configs enable it, we should follow that precedent.

TheIsotopes:
@SteveCEvans Please come down again,
what you have entered are the settings you would like (i.e. user settings), the manufacturer has not adopted these in the settings either.

if this is how you want it, then we also have to adapt the MatekF722Mini/MiniSE (UART6 = ESC) and MatekF405MINI (UART5 = ESC) to that effect. Just because it's in the manual doesn't mean it's included in the default settings.

blckmn:
I feel we need to consider a few things here.

First is what's best for the user as a default setting.
Second is what's easiest to maintain.
Third is what's the manufacturers preference.

Weighing up the above means we may have differences in standards between targets... but the best outcome for Users, Developers and the Manufacturers.

We also don't need to get it right always. We can learn and adjust.

SteveCEvans:
My take on this is simple. If we have a choice between "it's in the config and just works" or "the user needs to find the manual and figure it out themselves" then the preference is clear from the perspective of the user themselves, and those providing support, be that the manufacturer, or more likely the community support channels/forums.

In light of that my view on if we should declare, where possible, the function of a given pin, is that we should do so in the config file. Taking the manual in question here [MATEKH743-SLIM ](http://www.mateksys.com/?portfolio=h743-slim#tab-id-4) we can see that there is a connectors for ESC for which I'd say it's fair to say there's a default use for UART8 which the user can change if they really want to, but should get the sensible default. For the GPS unit though, wired to UART2 we could argue that that's just any old serial port and you could use any (although some of us would know that UART1 for UART3 would be bad choices as then DFU would stop working...) so maybe setting that to GPS would be going too far... or would it?


So lets gather some facts to help us (a bit).

With regard to any precedent of declaring port usage for the ESC sensor, searching for configs with `FUNCTION_ESC_SENSOR` declared as the port function (1024), we find there are 26.

```
$ grep "serial [0-9] 1024" configs/default/* | wc
      26     182    1952
```

Of which only 12 enable the feature. So where defined the precedent is (just) to not enable the feature.

```
$ grep "feature ESC_SENSOR" configs/default/* | wc
      12      24     693
```

I'll be honest, I find this surprising; I'm sure there will be more than 26 FCs which have a connector for their ESC.

We have 262 configs which define a LED_STRIP pin.

```
$ grep "resource LED_STRIP" configs/default/* | wc
     262    1047   17027
```

Of which only 53 enable the feature, so I think we can safely say the precedent is not to enable the feature.

```
$ grep "feature LED_STRIP" configs/default/* | wc
      53     106    3141
```

We have 66 FCs that enable TELEMETRY by default, yet all will be capable of it.

```
$ grep "feature TELEMETRY" configs/default/* | wc
      66     132    3882
```

Given the advent of presets perhaps we should just remove all feature statements from the config files and have these controlled instead by FC agnostic presets.

blckmn:
My thought is we only enable necessary features, or features that are 'built in' or 'on board'. 

Everything else is disabled.

TheIsotopes:
@blckmn I totally agree that just because one person wants it doesn't mean a thousand others want it too.

For example, if I had opened a PR, where I would like the UART5 from the MATEKF405MINI to become an ESC, then I could bet that the devs would ask - what's the point - and the PR would be closed without comment.

SteveCEvans:
> For example, if I had opened a PR, where I would like the UART5 from the MATEKF405MINI to become an ESC, then I could bet that the devs would ask - what's the point? - and the PR would be closed without comment.

You pick a contentious example! 😀UART5 is on the ESC connector so the manufacturers intent was clearly that it be used for that purpose, but it’s not a connector type that’s in common usage. Still, I’d argue that UART5 as the ESC SENSOR was the manufacturer’s intent and would support that as a default.

Look, I’m seriously not trying to be awkward here, but want stuff to be as easy as possible for the user.

blckmn:
@SteveCEvans in that example I'd enable it IF it was a RTF unified target or it was sold with the ESC as a package. I'd see that as absolutely clear intent. If it's up to the user to wire it up, I'd disable it by default, the user needs to verify their particular intent... by enabling it.

SteveCEvans:
@blckmn  Agreed, so the serial port definition would reflect the manufacturer’s intent that this be an ESC sensor. But the `feature ESC_SENSOR` would not be present in the config. I think we’re on the same page.

configs/default/MTKS-MATEKH743.config
```diff
@@ -159,12 +159,11 @@ dma pin B09 0
 # feature
 feature -RX_PARALLEL_PWM
 feature RX_SERIAL
-feature LED_STRIP
+feature TELEMETRY
 feature OSD
 
 # serial
 serial 5 64 115200 57600 0 115200
-serial 7 1024 115200 57600 0 115200
```
DusKing1, line 167: so this should be removed

DusKing1 (changes requested):

MATEKSYS (approved):

MATEKSYS:
> * removed User Settings
> * set feature to defaults
> 
> @SteveCEvans this reflects the configuration 100% to the legacy target.
> 
> @MATEKSYS can you confirm this.

Yes. We usually just enable Serial RX in ports tab.

SteveCEvans:
In that case I defer to the manufacturer.

SteveCEvans (approved):

PR #558, opened 2022-01-17
fix MATEKF405TE, MATEKH743 config

TheIsotopes:
Same reason as already discussed in #555.

**Edit:** I am still considering removing the setting `set serialrx_provider = xxxx` from all MTKS targets, which usually has to be adjusted by the user. What do the devs think of this?
Of course, the `MTKS-MATEKF411RX.config` is not affected.


haslinghuis (dismissed):

MATEKSYS:
set serialrx_provider = xxxx is needed

MATEKSYS (approved):

TheIsotopes:
@MATEKSYS ok, discarded. was just a thought.

TheIsotopes:
@MATEKSYS Overlooked entries `set serialrx_provider = SBUS` have been added.

MATEKSYS (commented):
we set it to CRSF on purpose in this target

MATEKSYS (approved):

MATEKSYS (approved):

PR #559, opened 2022-01-20
add axisflyingf7 target

FCtar:
Hi,

please add this target for a new FC from AxisFlying

KR


configs/default/AXFL-AXISFLYINGF7.config
```diff
@@ -0,0 +1,111 @@
...
+set dshot_burst = ON
+set current_meter = ADC
+set battery_meter = ADC
+set ibata_scale = 179
+set beeper_inversion = ON
+set beeper_od = OFF
+set max7456_spi_bus = 3
+set dashboard_i2c_bus = 1
+set flash_spi_bus = 1
+set gyro_1_spibus = 2
```
haslinghuis, line 111: Should we add
```
set gyro_1_bustype = SPI
```

haslinghuis (approved):

PR #580, opened 2022-04-03
Fix beeper on Diatone Mamba MK4 H743

haslinghuis:
Fixes https://github.com/betaflight/betaflight/issues/11499


uralm1:
What about `beeper_frequency = 0` ? Do someone use a pwm buzzer? I think pwm buzzer will not work - there is no timer for it. For me `beeper_frequency` wasn't zero by default.

uralm1:
> What about `beeper_frequency = 0` ? Do someone use a pwm buzzer? I think pwm buzzer will not work - there is no timer for it. For me `beeper_frequency` wasn't zero by default.

Forget. I see other configs don't include `beeper_frequency = 0` defaulting to active buzzer.

McGiverGim (approved):

blckmn (approved):

PR #599, opened 2022-05-07
Add New Foxeer F745 AIO V2 Target

Johnsnow-foxeer:
Would like to add Foxeer F745 AIO V2 target configuration based on BMI270 gyro, thank you!

### DIFF
--- a/configs/default/FOXE-FOXEERF745V2_AIO.config
+++ b/configs/default/FOXE-FOXEERF745V2_AIO.config
@@ -1,0 +1,107 @@
+# Betaflight / STM32F745 (S745) 4.1.0 Sep 25 2020 / 01:23:16 (064ca9b75) MSP API: 1.42
+
+board_name FOXEERF745V2_AIO
+manufacturer_id FOXE
+
+# resources
+resource BEEPER 1 D02
+resource MOTOR 1 B04
+resource MOTOR 2 B05
+resource MOTOR 3 B01
+resource MOTOR 4 B00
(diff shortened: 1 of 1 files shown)


haslinghuis (approved):

configs/default/FOXE-FOXEERF745V2_AIO.config
```diff
@@ -0,0 +1,107 @@
...
+# master
+set align_mag = CW180
+set mag_align_yaw = 1800
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set baro_bustype = I2C
+set baro_i2c_device = 1
+set serialrx_provider = SBUS
+set blackbox_device = SPIFLASH
+set motor_pwm_protocol = DSHOT600
```
Asizon, line 96: Is this an aio board with integrated ESC? If not this line should be removed.

Asizon (changes requested):

configs/default/FOXE-FOXEERF745V2_AIO.config
haslinghuis: @Asizon yes it is only another gyro being used with different orientation

Johnsnow-foxeer: > Is this an aio board with integrated ESC? If not this line should be removed.

Yes, ESC is integrated

Asizon (approved):

PR #609, opened 2022-05-17
Modify BETAFPV F722 target

betafpv-engineer:
Hi,all.

I request to modify the BETAFPVF722 target, this includes 3 places.

In order for the BETAFPVF722 target to use the dshot_bitbang function properly, I request to remove the CAMERA_CONTROL setting. In the flight controller that has been released, the CAMERA_CONTROL function has not been designed. In fact, it can complete the CAMERA_CONTROL function in motor5 or 6.

motor_pwm_protocol should use the default settings.

Added dual gyroscope settings.


haslinghuis (approved):

Asizon (approved):

Asizon:
Good to see 2 gyros here @betafpv-engineer! Both sharing same spi bus??

PR #611, opened 2022-05-21
Remove feature DYNAMIC_FILTER

haslinghuis:
After https://github.com/betaflight/betaflight/pull/10673 we need to remove feature `DYNAMIC_FILTER`


KarateBrot (approved):
Good find

MATEKSYS (approved):

Asizon:
Good, but what happens now if we want to use an old betaflight version? Aniway dynamic filter default is always enabled.

DusKing1 (approved):

PR #624, opened 2022-06-14
Add a new target TUNERCF405

TUNERC-Aria:
We'd like to add TUNERCF405 target file to the unified-targets repo.
This F405 AIO board is integrated with ESC. And the ESCs are pre-flashed with Bluejay.
The reason why we set motor_pwm_protocol = DSHOT600 is that the Bluejay doesn't support PWM.
Much appreciate your help in advance!

### DIFF
--- a/configs/default/TURC-TUNERCF405.config
+++ b/configs/default/TURC-TUNERCF405.config
@@ -1,0 +1,110 @@
+board_name TUNERCF405
+manufacturer_id TURC
+
+# resources
+resource BEEPER 1 B02
+resource MOTOR 1 B00
+resource MOTOR 2 A08
+resource MOTOR 3 B08
+resource MOTOR 4 A15
+resource MOTOR 5 C09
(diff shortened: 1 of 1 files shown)


haslinghuis (changes requested):
Please fix these problems:

1. PR is not a feature branch
2. CRLF line endings not allowed
3. No newline at EOF

TUNERC-Aria:
@haslinghuis  Thanks for your quick reply. 
About the second problem, I have fixed it by changing the line ending to LF. And about the last problem, I have added a new line at EOF. But I don't know how to fix the first problem " PR is not a feature branch ", could you please advise how to solve the problem? Thanks in advance.

haslinghuis:
Need to squash commits and we can't accept PR's based on master branch.
Not sure if this helps: https://gist.github.com/b-b3rn4rd/3e271b111b2fb92bffa2

PR #627, opened 2022-06-20
Add a new target TUNERCF405

TUNERC-Aria:
#624 is closed now. And the problem "PR is not a feature branch" has been fixed too.


blckmn (approved):

haslinghuis:
> This F405 AIO board is integrated with ESC. And the ESCs are pre-flashed with Bluejay.
The reason why we set motor_pwm_protocol = DSHOT600 is that the Bluejay doesn't support PWM.

haslinghuis (approved):

haslinghuis:
Made a new PR to test if anything wrong on our side.

https://github.com/betaflight/unified-targets/pull/629

PR #631, opened 2022-06-27
Update SJET-AOCODAF405.config

aocodarc:
The old configs can only be used on board aocodaf405 v1.0, and now we have A new board aocodaf405 v1.1.
The new board solves the S7 motor and S8 motor conflict problem.To do this, need to change the CS pins of OSD and gyro.
We're not going to create a new configuration file, so overwrite the old configuration file directly.
Can we do it this way?


daleckystepan:
If you change the config everyone with old board who updates will have your new config - so non properly working board. I think we are not able to do it this way - assuming board was publicly available.

haslinghuis:
New (board) revision needs V2 target.
`SJET-AOCODAF405V2`

aocodarc (commented):
yes,the changes are ok, TKS

PR #646, opened 2022-07-13
Added motor protocol setting for InfinityAIOV2 on-board ESCs

Romin0615:
Added motor protocol setting for InfinityAIOV2 on-board ESCs


haslinghuis (approved):

KarateBrot (approved):
That's a great idea

PR #652, opened 2022-07-14
Add HGLRCF722MINI config file

ASDosjani:
The original HGLRCF722 target was used by the mini board and this board has SPI baro instead of i2c so I decided to make a target for it so people don't have to use cli again to make their baro work.


haslinghuis:
@HGLRC-D please check.

HGLRC-D:
> @HGLRC-D please check.

Sorry, this is past F722 and mini version. We want it to be named "HGLRCF722E".

configs/default/HGLR-HGLRCF722E.config
```diff
@@ -0,0 +1,133 @@
...
+serial 0 64 115200 57600 0 115200
+
+# master
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set baro_bustype = SPI
+set baro_spi_device = 1
+set blackbox_device = SPIFLASH
+set dshot_burst = ON
+set motor_pwm_protocol = DSHOT600
```
haslinghuis, line 120: Please remove

haslinghuis (changes requested):
Please squash commits after making requested change.

configs/default/HGLR-HGLRCF722E.config
ASDosjani: I can't remove the "set baro_spi_device = 1" because the baro won't work without it. I removed the rest.

PR #654, opened 2022-07-15
Add HGLRCF722E target

ASDosjani:
The HGLRCF722 target was used by the mini board and this board has SPI baro instead of i2c so I decided to modify the original so people don't have to use cli again to make their baro work.
The discussions can be found in #652.


Asizon:
Duplicate of https://github.com/betaflight/unified-targets/pull/652
Please check opened PR, thanks

PR #667, opened 2022-08-26
Added IFLIGHT F411 RX SX1280 target

Guidus93:
Added IFLIGHT F411 ERLS RX target


configs/default/IFRC-IFLIGHT_F411SX1280.config
```diff
@@ -0,0 +1,84 @@
+# Betaflight / STM32F411SX1280 (S4SX) 4.3.1 Jul 13 2022 / 03:33:32 (8d4f005) MSP API: 1.44
+
+board_name IFLIGHT_F411SX1280
```
haslinghuis, line 3: I would even prefer the shorter name `F4SX1280` as it's the only unified target with SX12x0

Guidus93: Ok updated to F4SX1280

Asizon (approved):

PR #671, opened 2022-08-31
Rename and add AOCODAF4 targets

ASDosjani:
I have this FC but there was not target for it. I saved a dump and diff and compared it to the V2 target. It's almost the same, only the gyro alignment was different so that's why I'm making this PR. Now it's working perfectly.
@aocodarc please check this.

Edit: I talked to the manufacturer and it turned out that there a 2 revisions of the board. One is equipped with mpu6500, the other is mpu6000 but the gyro alignment is different so I get the permission from the manufacturer to rename the original V2 to V2MPU6500 and make a new target for the MPU6000.
The original V2 target is designed for the MPU6500 and V1.1 board which hopefully will be renamed to V2 according to the manufacturer.

### DIFF
--- a/configs/default/SJET-AOCODAF405V2MPU6000.config
+++ b/configs/default/SJET-AOCODAF405V2MPU6000.config
@@ -1,0 +1,125 @@
+# Betaflight / STM32F405 (S405) 4.2.11 Nov  9 2021 / 20:27:49 (948ba6339) MSP API: 1.43
+
+#define USE_BARO_BMP280
+#define USE_GYRO_SPI_MPU6000
+#define USE_ACC_SPI_MPU6000
+
+board_name AOCODAF405V2MPU6000
+manufacturer_id SJET
+
+# resources
+resource BEEPER 1 B08
+resource MOTOR 1 C06
+resource MOTOR 2 C07
+resource MOTOR 3 C08
+resource MOTOR 4 C09
+resource MOTOR 5 A15
+resource MOTOR 6 A08
+resource MOTOR 7 B10
+resource MOTOR 8 B11
+resource PPM 1 A03
+resource LED_STRIP 1 B01
+resource SERIAL_TX 1 A09
+resource SERIAL_TX 2 A02
+resource SERIAL_TX 3 C10
+resource SERIAL_TX 4 A00
+resource SERIAL_TX 5 C12
+resource SERIAL_RX 1 A10
+resource SERIAL_RX 2 A03
+resource SERIAL_RX 3 C11
+resource SERIAL_RX 4 A01
+resource SERIAL_RX 5 D02
+resource I2C_SCL 1 B06
+resource I2C_SDA 1 B07
+resource LED 1 C13
+resource SPI_SCK 1 A05
+resource SPI_SCK 2 B13
+resource SPI_SCK 3 B03
+resource SPI_MISO 1 A06
+resource SPI_MISO 2 B14
+resource SPI_MISO 3 B04
+resource SPI_MOSI 1 A07
+resource SPI_MOSI 2 B15
+resource SPI_MOSI 3 B05
+resource ESCSERIAL 1 C11
+resource ADC_BATT 1 C02
+resource ADC_RSSI 1 C03
+resource ADC_CURR 1 C01
+resource FLASH_CS 1 C00
+resource OSD_CS 1 A13
+resource GYRO_EXTI 1 C04
+resource GYRO_CS 1 A04
+resource USB_DETECT 1 B12
+
+# timer
+timer A03 AF2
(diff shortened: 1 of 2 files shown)


Asizon:
Hello, thought that no new target is needed if the only change is gyro aligment.

ASDosjani:
> Hello, I think that no new target is needed if the only change is gyro aligment.

Alright but than why the V2 is even exist? I didn't find a board for this target. If it is designed for this v1.1 than why the gyro alignment missing?

KarateBrot:
Adding it makes sense if there is a V1.1 out but it differs from V1.0

haslinghuis (dismissed):

haslinghuis:
We need to convert this to manufacturer presets with versioning in the new build system

ASDosjani:
I experienced in the meantime that `ibata_scale` is completely off. I think they just wrote some value completely random. The correct value is around 65 in my opinion. Is there any good way to determine this from the components? For example from resistor values if the FC use a voltage divider.

ASDosjani:
@KarateBrot I didn't found a single piece information about the V2 board. I think it was originally designed for my V1.1 but the naming changed. @aocodarc can you confirm this? I can make this PR to fix the gyro alignment only in the V2 config file.

ASDosjani:
I updated this PR, please reade first comment.

haslinghuis:
This PR is missing #defines for hardware being used.
Also don't include hardware (gyro) options in target naming.

ASDosjani:
> This PR is missing #defines for hardware being used. Also don't include hardware (gyro) options in target naming.

What do you recommend for the naming?
Are these #defines required for the new build system?

sugaarK:
> > This PR is missing #defines for hardware being used. Also don't include hardware (gyro) options in target naming.
> 
> What do you recommend for the naming? Are these #defines required for the new build system?

Yeah you do… 

Here is an example of the new defines..  

https://github.com/betaflight/unified-targets/blob/master/configs/default/AIRB-NOX.config

ASDosjani:
> Yeah you do…
> 
> Here is an example of the new defines..
> 
> https://github.com/betaflight/unified-targets/blob/master/configs/default/AIRB-NOX.config

Can I use #if to change the gyro rotation based on the gyroscope?

sugaarK:
> > Yeah you do…
> > Here is an example of the new defines..
> > https://github.com/betaflight/unified-targets/blob/master/configs/default/AIRB-NOX.config
> 
> Can I use #if to change the gyro rotation based on the gyroscope?

You’d define it like this 

set gyro_1_sensor_align = CW90FLIP
 set gyro_1_align_pitch = 1800
 set gyro_1_align_yaw = 900

sugaarK:
As you would in cli

ASDosjani:
> As you would in cli

Okay, but the problem is that there are two version of the board with two different gyro and their rotation is different, that's why I made two targets. Is it okay?

haslinghuis:
Yes targets with v1, v11, v2 should be okay if manufacturer uses the same revisions.

ASDosjani:
> Yes targets with v1, v11, v2 should be okay if manufacturer uses the same revisions.
> 
> The defines will allow to make use of the hardware in the configuration, so #else does not work here 🤔
> 
> As this is in final development stage we don't have any updated documentation yet.

I wasn't clear, sorry. There is no V2 board in existence however the target was made for the v1.1 hardware. The manufacturer don't want to make a v1.1 target instead he want to rename his hardware to v2. So the problem here is that this board can be ordered with mpu6000 or mpu6500 but the gyro alignment is different for this two but the FC is the same. So I made two targets with the correct alignment based on the used gyro. I will add the #define lines as soon as we clear this out.

haslinghuis:
aocoda should decide on versioning. If they don't want another version so be it. We have to direct support about it to aocoda.

ASDosjani:
>

ASDosjani:
Will this new build system be capable of running these custom cli command based on the definitions? Or how should I solve this two different gyro alignment?
I can give you Aocodarc availability if I understand the last sentence correctly.

haslinghuis:
No. The definitions only pull in required firmware parts for the target to build. So we need another version or use presets for the same board.

ASDosjani:
> No. The definitions only pull in required firmware parts for the target to build. So we need another version or use presets for the same board.

Okay, the presets are overkill in my opinion and not everyone use them so can I stick with this two target? I will add the #defines tomorrow.

ASDosjani:
@haslinghuis I added the #defines

haslinghuis (approved):

Asizon (changes requested):
@aocodarc what do you think about adding that targets? Thry are necessary?

Asizon (approved):

PR #679, opened 2022-09-28
Bug fixes

tobeornottobe81:
When two-way DSHOT is used, a timer conflict is prompted, To reduce the use of TIMER,  So cancel camera_ Control function.


sugaarK (approved):

haslinghuis:
@Guidus93 do you agree with this change?

Guidus93:
> @Guidus93 do you agree with this change?

No problem here

haslinghuis (approved):

PR #688, opened 2022-10-21
Add UART7_TX resource to JHEF745

alteman:
JHEMCU F745 FC missing uart7 TX pad, added it. No resource conflicts.


haslinghuis:
Please also add #defines for the target as we are migrating to a new build API.

For an example please look at https://github.com/betaflight/unified-targets/blob/master/configs/default/AIRB-NOX.config

blckmn (approved):

PR #696, opened 2022-10-26
TUNERCF405 config "Dshot_bitbang" is changed from "AUTO" to "ON".

TUNERC-Aria:
"Dshot_bitbang" is changed from "AUTO" to "ON". This change is made to solve the issue that when Bidirectional Dshot is disabled, the LED strip would also be disabled, which would result in the LED not functioning.

### DIFF
--- a/configs/default/TURC-TUNERCF405.config
+++ b/configs/default/TURC-TUNERCF405.config
@@ -95,6 +95,7 @@ feature OSD
 set serialrx_provider = CRSF
 set blackbox_device = SPIFLASH
 set dshot_burst = ON
+set dshot_bitbang = ON
 set motor_pwm_protocol = DSHOT600
 set current_meter = ADC
 set battery_meter = ADC
--- a/configs/default/TURC-TUNERCF405.config
+++ b/configs/default/TURC-TUNERCF405.config
@@ -1,5 +1,12 @@
 # Betaflight / STM32F405 (S405) 4.3.0 Jun 14 2022 / 00:47:24 (229ac66) MSP API: 1.44
 
+#define USE_ACCGYRO_BMI270
+#define USE_BARO_BMP280
+#define USE_BARO_SPI_BMP280
+#define USE_BARO_DPS310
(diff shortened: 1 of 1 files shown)


sugaarK:
Please update the target to include this information to suit our new build platform.


2.3. Add #defines for hardware drivers

edit the file from the previous step, and add #defines for accelerometer, barometer, gyro and magnetometer hardware drivers to be included for the cloud build API.
For example on [this target](https://github.com/betaflight/unified-targets/blob/master/configs/default/AIRB-NOX.config) it will look like:

# Betaflight / STM32F411 (S411) 4.1.0 Oct 16 2019 / 11:57:34 (c37a7c91a) MSP API: 1.42

#define USE_BARO_SPI_BMP280
#define USE_BARO_BMP280
#define USE_GYRO_SPI_MPU6000
#define USE_GYRO_SPI_MPU6500
#define USE_ACC_SPI_MPU6000
#define USE_ACC_SPI_MPU6500

board_name NOX
manufacturer_id AIRB

TUNERC-Aria:
> 

Thank you for the heads up.
Before I edit the file, I want to confirm a few details with you.
#1 The TUNERC Poly F405 board has onboard BMI270. So I should add the following two lines in the file, right?
#define USE_GYRO_SPI_BMI270
#define USE_ACC_SPI_BMI270

#2 We don't have a barometer onboard but we have IIC connection ports on the aio flight controller. Without defining the barometer in the file, can the users still get the barometer to work properly? 

#3 Should I also add SPI Flash and OSD in the edited file?

sugaarK:
Only need to add the defines you need so.. yes to flash no to baro unless you think they might need it. Osd is an option in the build system so no need to worry about that

TUNERC-Aria:
Thank you for your quick response.
I have added the defines for the hardware drivers in the edited file.
Please check it out. Thank you for your review in advance.

haslinghuis (approved):

sugaarK (approved):

PR #702, opened 2022-11-02
Diatone H743: improve mag i2c settings, add #define properties

PhilippMolitor:
Seems like the I2C bus that is exposed on the PCB is `1` and not `0`, also the mag device needs to be set to `I2C` in order to work plug-and-play like with the usual hardware we use.
I tested this on my hardware successfully with the Matek M10-5883 which has the QMC5883 mag.

### DIFF
--- a/configs/default/DIAT-MAMBAH743.config
+++ b/configs/default/DIAT-MAMBAH743.config
@@ -141,6 +141,8 @@ serial 7 0 115200 57600 0 115200
 
 set baro_bustype = I2C
 set baro_i2c_device = 1
+set mag_bustype = I2C
+set mag_i2c_device = 1
 set sdcard_mode = OFF
 set flash_spi_bus = 3
 set blackbox_device = SPIFLASH
--- a/configs/default/DIAT-MAMBAH743.config
+++ b/configs/default/DIAT-MAMBAH743.config
@@ -3,6 +3,16 @@
 board_name MAMBAH743
 manufacturer_id DIAT
 
+#define USE_GYRO_SPI_MPU6000
+#define USE_ACC_SPI_MPU6000
+#define USE_ACCGYRO_BMI270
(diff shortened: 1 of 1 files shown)


haslinghuis:
Could you also add new defines to the target:
https://github.com/betaflight/betaflight/blob/master/docs/TargetMaintenance/CreatingAUnifiedTarget.md#23-add-defines-for-hardware-drivers

Also please add #define MAX7456 as it's not in the documentation yet.

PhilippMolitor:
Do you think adding Gyro (and maybe other) defines as well makes sense? I am not really too fond of the cloud build API, but it seems like they also belong there. Could do all of that in one go :)

haslinghuis:
Yes you are right it's `#define USE_MAX7456`

As there is no list yet, for now we have to look at other targets for examples for this case. (Aikon F4 or Iflight H743 might be good examples)

PhilippMolitor:
Alright, maybe some questions the docs should cover as a future reference (also because I don't know where else I could put them):
- What defines are available (maybe grouped by device type?)
- If a target is available in  multiple revisions with variations on the sensors, how should that be handled?
- The iFlight target for example has mags listed, which are not on board for sure. Should I also add expected sensors (like the mag) as well?

PhilippMolitor:
I added the properties for the known IMUs, baro, OSD chip as well as the same mag list as in the iFlight H7 target. I sadly don't know anything about the flash chip, so I left this one out. (It is some 128MB / 1GBit flash)

haslinghuis:
No please only add drivers which are available on a board revision. If there is no magnetometer we don't need mag drivers.
We need to publish a list soon to communicate to manufacturers

haslinghuis:
We can try `#define USE_FLASH_M25P16`

haslinghuis (approved):

ctzsnooze (approved):

PR #712, opened 2022-11-08
Update Diatone targets

haslinghuis:


blckmn (approved):

sugaarK (approved):

PhilippMolitor:
I noticed these boards had a significant target renaming happening - my H743 (now _2022A) is not able to auto-detect the target for firmware upgrades now.
Is there anything in place to mitigate this? Either a notification, aliases or something alike? If not, is there a simple way to force a new board name via the CLI?

haslinghuis:
@PhilippMolitor was trying to be consistent with target naming https://www.diatone.us/products/mamba-mk4-h743-v2-flight-control-30mm-m3

You can override the target name while flashing. But I'm considering to revert the name change.

PhilippMolitor:
Yeah, that's kind of what i was thinking... my thoughts:

- Naming the targets `_2022A` and `_2022B` is very confusing, as the manufacturer gave us no clue that the "old" FC is the `_2022A` and also removed all documentation from their shop. In their Support/Docs website there is documentation/cli dumps for the H743 (legacy) target, but no mention of `_2022A` and `_2022B.`
- I am thinking of `MAMBAH743` as the target name for the v1, which is the old one as it was, and `MAMBAH743_V2` as the new one is mostly listed as "V2".
- If a `_V2` suffix is not suitable, then it should probably stay with `_2022B` for compatability with already shipped FCs.
- The manufacturer should be notified that the migration from no suffix to `_2022A` was not communicated and breaks updating for owners of that revision.

haslinghuis:
Yes I made the change as other boards were using the same extension in their name.
Going to revert to `MAMBAH743`
The V2 has ICM42605 instead of MPU6000 or BMI270 but there is no problem combining these into a single target to mitigate confusion.

PhilippMolitor:
Yeah, the targets are virtually identical... the only exception relevant to the actual target config (not regarding cloud build) I see is that on UART4 there is the WiFi chip for V2. (even though I didn't get it to work so far)

haslinghuis:
Will leave it to Diatone to submit a new configuration for V2

Have pushed a new merge request. Thanks for the heads up.

PR #713, opened 2022-11-09
Update HAMO-CRAZYBEEF4SX1280.config

blckmn:
Adding flash support. Noting this target is used by 

https://www.happymodel.cn/index.php/2022/06/11/pancake-2-6s-aio-f4-flight-controller-built-in-spi-elrs-2-4g-and-400mw-openvtx-compatible-with-20x20-and-35-5x35-5-stack/21-7/

Which may need its own target if this flash chip is not standard on CRAZYBEE boards.

### DIFF
--- a/configs/default/HAMO-CRAZYBEEF4SX1280.config
+++ b/configs/default/HAMO-CRAZYBEEF4SX1280.config
@@ -12,6 +12,8 @@
 #define USE_RX_EXPRESSLRS_TELEMETRY
 #define USE_RX_SX1280
 #define RX_CHANNELS_AETR
+#define USE_FLASH_W25M
+#define USE_FLASH_W25Q128FV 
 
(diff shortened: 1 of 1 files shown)


ASDosjani:
> Adding flash support. Noting this target is used by
> 
> https://www.happymodel.cn/index.php/2022/06/11/pancake-2-6s-aio-f4-flight-controller-built-in-spi-elrs-2-4g-and-400mw-openvtx-compatible-with-20x20-and-35-5x35-5-stack/21-7/
> 
> Which may need its own target if this flash chip is not standard on CRAZYBEE boards.

Why we need #define USE_FLASH_W25M?
I only see a 25Q128 flash.

haslinghuis (approved):

sugaarK (approved):

PR #714, opened 2022-11-09
add some more defines

ASDosjani:
I'm not sure  about FLYWOOF745 and HAKRCF722D baro. These manufacturer descriptions are horrible. They only write something like this, barometer: yes... very specific

P.s: I won't have time to add more defines in the future.


ASDosjani:
I recommend the Reverse Search extension for VSCode to find the files which doesn't contains `#define`

sugaarK (approved):

PR #730, opened 2022-11-13
Fix file endings

ASDosjani:
I added the `fix-file-endings.sh` script which grants that config files only have one empty line at their ends. Only problem is that git doesn't show the config files correctly. If I use the script the files and the git differences in VSCode seems correct but after I push my changes it looks like it removed the file endings. Any idea how to solve this?


KarateBrot:
I don't see anything wrong with it. Looks perfectly normal. You added a line feed but did not start a new line, so Github correctly displays the last line as the last line (and not a new line like in a text editor).

KarateBrot (approved):

ASDosjani:
> I don't see anything wrong with it. Looks perfectly normal. Nice job. You added a line feed but did not start a new line, so Github correctly displays the last line as the last line (and doesn't display a new line, like a text editor would).
> 
> Editors need to do this so that you can set the cursor to a new line (which technically doesn't exist before you start typing). It's just a display thing (it's a "fake empty line").

Wow, thank you very much. So there's nothing wrong here.

haslinghuis (approved):

KarateBrot:
I reckon there was often an additional line at the end that only consisted of one LF, e.g.
```
blabla\LF
blablabla\LF
\LF
```
I even encountered one instance where there were two empty lines at the end, e.g.
```
blabla\LF
blablabla\LF
\LF
\LF
```
and your script made them look like this
```
blabla\LF
blablabla\LF
```

ASDosjani:
> I reckon there was often an additional line at the end that only consisted of one LF, e.g.
> 
> ```
> blabla\LF
> blablabla\LF
> \LF
> ```
> 
> While reviewing your PR I even encountered one instance where there were two "empty lines" at the end, e.g.
> 
> ```
> blabla\LF
> blablabla\LF
> \LF
> \LF
> ```
> 
> and your script made them look like this
> 
> ```
> blabla\LF
> blablabla\LF
> ```

Yes, this was intended

PR #732, opened 2022-11-13
Default defines for gyro  / acc for DIAT-MAMBAF405US_I2C

blckmn:
These maybe incorrect


blckmn:
Already done! ;) thanks @haslinghuis

PR #735, opened 2022-11-15
Added GFPV-GEMEF411 target

GEME-Cyi:
Hi, Added GFPV GEMEF411 unif target (include hw define)


KarateBrot (dismissed):

configs/default/GFPV-GEMEF411.config
```diff
@@ -0,0 +1,123 @@
+# Betaflight / STM32F411 (S411) 4.3.0 Jun 14 2022 / 00:48:04 (229ac66) MSP API: 1.44
+
+#define USE_GYRO_SPI_ICM42688P
+#define USE_ACC_SPI_ICM42688P
+#define USE_ACCGYRO_BMI270
+#define USE_BARO_DPS310
+#define USE_BARO_BMP280
```
haslinghuis, line 7: Please remove BARO drivers.

We have to update documentation as BARO and MAG are now defined in the build process instead.

KarateBrot: But isn't this still useful information in case we want to switch to individual component defines at some point? Can we just ignore defines in the config file somehow if they are already defined by the compiler? @blckmn

blckmn: Happy for it to be a comment in the file, e.g. ## Board includes: BMP280 and DPS310 baros.

blckmn: Alternatively I could create an ignore list of defines in the processor... The problem is there were many files with EVERY baro listed, which makes no sense. I'd be ok with only those that were actually physically on the board.

GEME-Cyi: OK. I have removed the baro defines.

haslinghuis (approved):

KarateBrot (approved):

PR #739, opened 2022-11-19
Fix MAMBAF722_2022 (mk4) UART VTX port

tfranzel:
This almost gave me a heart attack thinking the new Vista is broken. :laughing: 

Turns out the serial port is wrong. It is UART3 instead of UART4. After that the Vista connection worked. The wiring diagram agrees.

I only changed `MAMBAF722_2022A` and `B` and not base `DIAT-MAMBAF722` as I am unsure about previous revisions. 

Official Wiring Diagram:
https://cdn.shopifycdn.net/s/files/1/0027/2708/4144/files/MK4_F722_MINI_42688P_-01.jpg

### DIFF
--- a/configs/default/DIAT-MAMBAF722_2022A.config
+++ b/configs/default/DIAT-MAMBAF722_2022A.config
@@ -107,7 +107,7 @@ feature OSD
 
 # serial
 serial 0 64 115200 57600 0 115200
-serial 3 1 115200 57600 0 115200
+serial 2 1 115200 57600 0 115200
 
 # EXTRAS
 resource PINIO 1 C02
--- a/configs/default/DIAT-MAMBAF722_2022B.config
+++ b/configs/default/DIAT-MAMBAF722_2022B.config
@@ -111,7 +111,7 @@ feature OSD
 
 # serial
 serial 0 64 115200 57600 0 115200
-serial 3 1 115200 57600 0 115200
+serial 2 1 115200 57600 0 115200
 
 # master
 set mag_bustype = I2C


configs/default/DIAT-MAMBAF722_2022A.config
```diff
@@ -107,7 +107,8 @@ feature OSD
 
 # serial
 serial 0 64 115200 57600 0 115200
-serial 3 1 115200 57600 0 115200
+serial 2 1 115200 57600 0 115200
+serial 3 0 115200 57600 0 115200
```
haslinghuis, line 111: Remove this line as this is default.

configs/default/DIAT-MAMBAF722_2022B.config
```diff
@@ -111,7 +111,8 @@ feature OSD
 
 # serial
 serial 0 64 115200 57600 0 115200
-serial 3 1 115200 57600 0 115200
+serial 2 1 115200 57600 0 115200
+serial 3 0 115200 57600 0 115200
```
haslinghuis, line 115: Remove this line as this is default.

tfranzel: which line? you mean 115 (`serial 3 0 115200 57600 0 115200`)?

I did that because I was under the impression that this config is layered on top of the base config

https://github.com/betaflight/unified-targets/blob/master/configs/default/DIAT-MAMBAF722.config

and there it gets enabled

https://github.com/betaflight/unified-targets/blob/6a788a3aa76a2daab7241e5b9090b724a75d316e/configs/default/DIAT-MAMBAF722.config#L111

As far as I understand, without line 115 I would have 2 ports (`serial 2` and `serial 3`) enabled, which would be wrong. Am I mistaken here?

haslinghuis: `DIAT-MAMBAF722` is not a base target for the other ones.
Line 115 is not needed anymore.

tfranzel: done. thanks for the clarification.

haslinghuis (approved):

sugaarK (approved):

PR #743, opened 2022-12-10
Create HARC-HAKRCF405V2.config

sugaarK:
new HARC target with ICM 42688-p gyro

### DIFF
--- a/configs/default/HARC-HAKRCF405ICM.config
+++ b/configs/default/HARC-HAKRCF405ICM.config
@@ -1,0 +1,129 @@
+# Betaflight / STM32F405 (S405) 4.2.4
+
+#define USE_GYRO_SPI_ICM42688P 
+#define USE_ACC_SPI_ICM42688P
+#define USE_FLASH_W25Q128FV
+#define USE_MAX7456
+
+board_name HAKRCF40ICM
+manufacturer_id HARC
+
+# resources
+resource BEEPER 1 C13
+resource MOTOR 1 C08
+resource MOTOR 2 C09
+resource MOTOR 3 A08
+resource MOTOR 4 A09
+resource MOTOR 5 A10
+resource MOTOR 6 B04
+resource MOTOR 7 B00
+resource MOTOR 8 B01
+resource LED_STRIP 1 B03
+resource SERIAL_TX 1 B06
+resource SERIAL_TX 2 A02
+resource SERIAL_TX 3 B10
+resource SERIAL_TX 4 A00
+resource SERIAL_TX 5 C12
+resource SERIAL_TX 6 C06
+resource SERIAL_RX 1 B07
+resource SERIAL_RX 2 A03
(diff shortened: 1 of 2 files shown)


configs/default/HARC-HAKRCF405ICM.config
```diff
@@ -0,0 +1,129 @@
...
+dma pin B01 0
+# pin B01: DMA1 Stream 2 Channel 5
+dma pin B08 0
+# pin B08: DMA1 Stream 7 Channel 2
+dma pin B03 0
+# pin B03: DMA1 Stream 6 Channel 3
+
+# feature
+feature RX_SERIAL 
+feature OSD
```
howels, line 108: OSD is user defined in cloud builds.

howels (changes requested):

configs/default/HARC-HAKRCF405ICM.config
```diff
@@ -0,0 +1,129 @@
+# Betaflight / STM32F405 (S405) 4.2.4
+
+#define USE_GYRO_SPI_ICM42688P 
+#define USE_ACC_SPI_ICM42688P
+#define USE_FLASH_W25Q128FV
+#define USE_MAX7456
+
+board_name HAKRCF40ICM
```
KarateBrot, line 8:
```suggestion
board_name HAKRCF405ICM
```

Are you sure this is a good board name? If in future there are other gyros on board this name will be very misleading.

KarateBrot (changes requested):

sugaarK:
What would you recommend @KarateBrot ? This came from my contact.. I asked what they wanted as when I got to it I’d be drinking and my go for something witty

configs/default/HARC-HAKRCF405ICM.config
sugaarK: This turns it on as a feature not as a define ? With out thia it won’t be be on when you flash the fc.. sorry was copying other targets

haslinghuis: #define USE_OSD does not enable feature OSD so this needs to stay.

haslinghuis:
Would suggest HARC-HAKRCF405V2 or add the two #defines to the existing config.

configs/default/HARC-HAKRCF405ICM.config
howels: Ok, thought this was defined centrally instead of in each board definition

sugaarK:
All fixed up thanks @haslinghuis

KarateBrot (approved):

haslinghuis:
Some fixes made:

Updated version to 4.4.0
Removed = from # dma
Removed `feature RX_SERIAL` as this is default (https://github.com/betaflight/betaflight/pull/11708)

blckmn (approved):

PR #758, opened 2022-12-26
SPRacingH7EXTREME - First cut of a unified target.

hydra:
Requires https://github.com/betaflight/betaflight/pull/12091

### DIFF
--- a/configs/default/SPRO-SPRACINGH7EXTREME.config
+++ b/configs/default/SPRO-SPRACINGH7EXTREME.config
@@ -1,0 +1,231 @@
+# Betaflight / STM32H750 (SP7E) 4.4.0 Dec 26 2022 / 12:28:11 (7671c7eb01) MSP API: 1.45
+
+#define TARGET_BOARD_IDENTIFIER "SP7E"
+#define USBD_PRODUCT_STRING "SPRacingH7EXTREME"
+
+#
+# Settings required to boot
+#
+# Note: hardware for all possible config storage subsystems needs to be known and initialised BEFORE loading
+# a config from it can work. (SPI/QSPI/SDCARD).
+#
+
+#define EEPROM_SIZE=8192
+
+#define USE_SPRACING_PERSISTENT_RTC_WORKAROUND
+
+#define USE_BUTTONS
+#define BUTTON_A_PIN            PE4
+#define BUTTON_A_PIN_INVERTED // Active high
+#define BUTTON_B_PIN            PE4
+#define BUTTON_B_PIN_INVERTED // Active high
+
+#define USE_QUADSPI
+#define USE_QUADSPI_DEVICE_1
+#define QUADSPI1_SCK_PIN PB2
+#define QUADSPI1_BK1_IO0_PIN PD11
+#define QUADSPI1_BK1_IO1_PIN PD12
+#define QUADSPI1_BK1_IO2_PIN PE2
+#define QUADSPI1_BK1_IO3_PIN PD13
+#define QUADSPI1_BK1_CS_PIN PB10
+#define QUADSPI1_BK2_IO0_PIN PE7
+#define QUADSPI1_BK2_IO1_PIN PE8
+#define QUADSPI1_BK2_IO2_PIN PE9
+#define QUADSPI1_BK2_IO3_PIN PE10
+#define QUADSPI1_BK2_CS_PIN NONE
+#define QUADSPI1_MODE QUADSPI_MODE_BK1_ONLY
+#define QUADSPI1_CS_FLAGS (QUADSPI_BK1_CS_HARDWARE | QUADSPI_BK2_CS_NONE | QUADSPI_CS_MODE_LINKED)
+
+#define USE_SPI
+
+#define USE_SPI_DEVICE_2
+#define SPI2_SCK_PIN            PD3
+#define SPI2_MISO_PIN           PC2
+#define SPI2_MOSI_PIN           PC3
+#define SPI2_NSS_PIN            PB12
+
+#define USE_SPI_DEVICE_3
+#define SPI3_SCK_PIN            PB3
+#define SPI3_MISO_PIN           PB4
+#define SPI3_MOSI_PIN           PD6
+#define SPI3_NSS_PIN            PA15
+
+#define USE_SPI_DEVICE_4
+#define SPI4_SCK_PIN            PE12
+#define SPI4_MISO_PIN           PE13
+#define SPI4_MOSI_PIN           PE14
+#define SPI4_NSS_PIN            PE11
+
+#define USE_FLASH_TOOLS
+#define USE_FLASH_W25N01G
+#define FLASH_QUADSPI_INSTANCE    QUADSPI
+#define CONFIG_IN_EXTERNAL_FLASH
+
+#define USE_SDCARD
+#define SDCARD_DETECT_PIN PD10
+#define SDCARD_DETECT_INVERTED
+#define SDIO_DEVICE             SDIODEV_1
+#define SDIO_USE_4BIT           true
+#define SDIO_CK_PIN             PC12
+#define SDIO_CMD_PIN            PD2
+#define SDIO_D0_PIN             PC8
+#define SDIO_D1_PIN             PC9
+#define SDIO_D2_PIN             PC10
+#define SDIO_D3_PIN             PC11
+
+#define TARGET_IO_PORTA 0xffff
+#define TARGET_IO_PORTB 0xffff
+#define TARGET_IO_PORTC 0xffff
+#define TARGET_IO_PORTD 0xffff
+#define TARGET_IO_PORTE 0xffff
+#define TARGET_IO_PORTF 0xffff
+#define TARGET_IO_PORTG 0xffff
+
+#
+# Settings to enable/configure remaining hardware
+#
+
+#define USE_USB_ID
+
+#define USE_I2C
+#define USE_I2C_DEVICE_1
+#define I2C1_SCL PB8
+#define I2C1_SDA PB9
+#define I2C_DEVICE (I2CDEV_1)
+
+
+#define USE_ACC
+#define USE_ACC_SPI_MPU6500
+#define USE_GYRO
+#define USE_GYRO_SPI_MPU6500
+#define ENSURE_MPU_DATA_READY_IS_LOW
+
+#define USE_PID_AUDIO
+
+#define VTX_RTC6705_OPTIONAL
+
+#define ADC1_DMA_OPT 8
+#define ADC3_DMA_OPT 9
+
+#define USE_FLASHFS
+#define ENABLE_BLACKBOX_LOGGING_ON_SDCARD_BY_DEFAULT
+
+#
+# Settings that are currently defined in target/common_pre.h for non-cloud builds that probably shouldn't be.
+# There are here to illustrate that they SHOULD be included in THIS target when they are removed by default.
+#
+
+#define _USE_MAG
+#define _USE_MAG_HMC5883
+#define _USE_MAG_QMC5883
+#define _USE_BARO
+#define _USE_BARO_BMP388
+#define _USE_TRANSPONDER
+#define _USE_OSD
+#define _USE_MAX7456
+#define _USE_VTX_COMMON
+#define _USE_VTX_CONTROL
+#define _USE_VTX_RTC6705_SOFTSPI
+
+#
+# Settings usually defined in common_pre.h for non-cloud-builds
+# 
+
+
+
+board_name SPRACINGH7EXTREME
+manufacturer_id SPRO
+
+resource BEEPER 1 D07
+resource MOTOR 1 A00
+resource MOTOR 2 A01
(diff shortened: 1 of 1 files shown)


hydra:
Testing:

1) use the bootloader to erase the config partition, see the SPRacingH7EXTREME manual.

Then either: 

2.a) flash attached binaries
[SPRACING-20221226-1637-4.4.0-SPRACINGH7EXTREME.zip](https://github.com/betaflight/unified-targets/files/10304017/SPRACING-20221226-1637-4.4.0-SPRACINGH7EXTREME.zip)
Use the `.hex` or `.bin` to firmware in the usual manner, either via configurator or via dfu-util.

2.b) run the build script from PR https://github.com/betaflight/betaflight/pull/12091
```
support/scripts/build_spracingh7extreme.sh
```
then flash the `obj/STM32H750.bin` or `obj/STM32H750.hex`.
```bash
; boot into spracing DFU bootloader mode first, see manual.
dfu-util.exe -a 0 -D obj/betaflight_4.4.0_STM32H750.bin -s 0x97ce0000:0x70000
```

3) then unplug, replug, firmware should boot, but no gyro will be detected due to missing resource configuration.

4) use configurator to access the CLI and paste in the contents of the `SPRO-SPRACINGH7EXTREME.config` file (in this PR and attached).
then issue `save` in the CLI.

5) observe that both gyros should be detected and you can continue testing other aspects of the firmware.

configs/default/SPRO-SPRACINGH7EXTREME.config
```diff
@@ -0,0 +1,225 @@
...
+#define SDIO_DEVICE             SDIODEV_1
+#define SDIO_USE_4BIT           true
+#define SDIO_CK_PIN             PC12
+#define SDIO_CMD_PIN            PD2
+#define SDIO_D0_PIN             PC8
+#define SDIO_D1_PIN             PC9
+#define SDIO_D2_PIN             PC10
+#define SDIO_D3_PIN             PC11
+
+#define TARGET_IO_PORTA 0xffff
```
blckmn, line 76: These are already defined in the target.h

configs/default/SPRO-SPRACINGH7EXTREME.config
```diff
@@ -0,0 +1,225 @@
...
+#define TARGET_IO_PORTF 0xffff
+#define TARGET_IO_PORTG 0xffff
+
+#
+# Settings to enable/configure remaining hardware
+#
+
+#define USE_USB_ID
+
+#define USE_I2C
```
blckmn, line 90: This is unnecessary? Should be in the resource management below.

configs/default/SPRO-SPRACINGH7EXTREME.config
```diff
@@ -0,0 +1,225 @@
...
+#define USE_I2C_DEVICE_1
+#define I2C1_SCL PB8
+#define I2C1_SDA PB9
+#define I2C_DEVICE (I2CDEV_1)
+
+#define _USE_MAG
```
blckmn, line 96: These are superfluous, assumed to be disabled, should be cleaned up.

configs/default/SPRO-SPRACINGH7EXTREME.config
```diff
@@ -0,0 +1,225 @@
...
+#define _USE_BARO
+#define _USE_BARO_BMP388
+
+#define USE_ACC
+#define USE_ACC_SPI_MPU6500
+#define USE_GYRO
+#define USE_GYRO_SPI_MPU6500
+#define ENSURE_MPU_DATA_READY_IS_LOW
+
+#define USE_PID_AUDIO
```
blckmn, line 109: This is a user option. Can be added to the options a user can add in their build.

configs/default/SPRO-SPRACINGH7EXTREME.config
```diff
@@ -0,0 +1,225 @@
...
+#define _USE_TRANSPONDER
+#define _USE_OSD
+#define _USE_MAX7456
+#define MAX7456_SPI_INSTANCE SPI4
```
blckmn, line 113: Resource mapping

configs/default/SPRO-SPRACINGH7EXTREME.config
```diff
@@ -0,0 +1,225 @@
...
+#define _USE_VTX_COMMON
+#define _USE_VTX_CONTROL
+#define _USE_VTX_RTC6705_SOFTSPI
+#define VTX_RTC6705_OPTIONAL
+
+#define ADC1_DMA_OPT 8
+#define ADC3_DMA_OPT 9
+
+#define USE_FLASHFS
+#define ENABLE_BLACKBOX_LOGGING_ON_SDCARD_BY_DEFAULT
```
blckmn, line 124: Custom defaults please.

configs/default/SPRO-SPRACINGH7EXTREME.config
```diff
@@ -0,0 +1,225 @@
...
+#
+# Note: hardware for all possible config storage subsystems needs to be known and initialised BEFORE loading
+# a config from it can work. (SPI/QSPI/SDCARD).
+#
+
+#define EEPROM_SIZE=8192
+
+#define USE_SPRACING_PERSISTENT_RTC_WORKAROUND
+
+#define USE_BUTTONS
```
blckmn, line 17: What are the buttons for? should they be using resource mapping?

blckmn:
It might be worth storing custom defaults in a page as part of the 128K flash - then the paste into cli is not needed and users can safely reset to defaults.

Alternatively we need to find a way to consider "custom defaults" active with some configuration stored elsewhere and flashed as part of the hex (configurator injects custom defaults into the hex).

configs/default/SPRO-SPRACINGH7EXTREME.config
hydra: see comments elsewhere regarding this.

configs/default/SPRO-SPRACINGH7EXTREME.config
hydra: see comments elsewhere regarding this.

configs/default/SPRO-SPRACINGH7EXTREME.config
hydra: it's not a user option, the audio hardware needs to be initialised.  enabling the pid audio FEATURE however IS a user option.

configs/default/SPRO-SPRACINGH7EXTREME.config
hydra: yes, that one I can move, done.

configs/default/SPRO-SPRACINGH7EXTREME.config
hydra: yes, and they should not be as targets need to be able to overwrite them, for example the H7RF uses specific TARGET_IO_PORTX settings to ensure that BF doesn't undo the OCTOSPI settings which would cause a hard-fault if it did.

configs/default/SPRO-SPRACINGH7EXTREME.config
hydra: there is no custom default support for H750/H730, not in the linker scripts and we certainly wouldn't want to waste more of the 448K RAM for a copy of some defaults.  There is no bootloader support for a custom-defaults flash partition. There is no BF support for a custom-defaults flash partition.  Happy to discuss adding H750/H730 custom defaults support in another issue, but this PR isn't the place.

configs/default/SPRO-SPRACINGH7EXTREME.config
hydra: buttons are used for various pre-config-loaded activities such as resetting the config, USB MSC and also for VTX band/channel/power settings.  See the main BF codebase: `init.c`, `vtx_control.c`, `pg\usb.c`, etc.

hydra:
remember that for non-memory mapped EXST targets the firmware blob is copies from external flash to RAM.

> It might be worth storing custom defaults in a page as part of the 128K flash - then the paste into cli is not needed and users can safely reset to defaults.

as noted elsewhere, there's only one flash sector on the H750/H730 targets and that's where the bootloader lives.  there's no chance to erase anything other than a single flash sector, also flashing both internal and external flash isn't supported anyway, either by the bootloader, as the bootloader exposes external flash via dfu only exactly because there's no sectors on the internal flash that can be erased by the user while the bootloader itself is running from internal flash, or by the configurator, or by any current distribution format (hex/bin). file hex files can contain non-contiguous blocks, the configurator would have to know what the blocks were for.  I've mentioned that the distribution format of BF could be changed to be a .json file containing build information and an array of 'image' objects that contain descriptions and a uuencoded gzip compressed blobs which the configurator could flash to according.  similar, but better, to how PX4 does it.  that way you could distribute a `.bf` json file with custom defaults blob and start address, firmware blob and start address, and any other blobs, and the configurator could flash these blobs to the start address and optionally the device (internal flash, external flash, etc) as required.   For example: https://jsoneditoronline.org/#left=cloud.0a3a37d4af614c0288df47388c750f1c

> Alternatively we need to find a way to consider "custom defaults" active with some configuration stored elsewhere and flashed as part of the hex (configurator injects custom defaults into the hex).

and also I'll repeat my comment from above here:

there is no custom default support for H750/H730, not in the linker scripts and we certainly wouldn't want to waste more of the 448K RAM for a copy of some defaults. There is no bootloader support for a custom-defaults flash partition. There is no BF support for a custom-defaults flash partition. Happy to discuss adding H750/H730 custom defaults support in another issue, but this PR isn't the place.

hydra:
@blckmn thanks for the review.  I replied to all comments and updated the PR.

configs/default/SPRO-SPRACINGH7EXTREME.config
blckmn: So the point is still valid, these should be a user option gate. So the user can chose to enable them, in the build, and then apply resource mapping to configure them.

configs/default/SPRO-SPRACINGH7EXTREME.config
blckmn: Ok for clarity this should be a user Build option, ie do I want to include this feature at all or not. Whether I run time include it is a different story. Of course it is your target config so if you want to force users to include it... happy days.

configs/default/SPRO-SPRACINGH7EXTREME.config
hydra: can you explain what you mean by 'user gate' and how to create one.  I don't recall seeing any documentation or use of that term until now.

If the user wants to enable buttons then the expectation would be that they always work, and not 'only work after config is loaded' so with that in mind they /have/ to be compile-time options in the case of the H750 targets.

configs/default/SPRO-SPRACINGH7EXTREME.config
hydra: no. the pin connected to the audio amp must be configured as ANALOG_OUT.  At all times.  It is not acceptable and bad for the on-board audio hardware if the pin for instance is set to any other GPIO mode.   It NEEDs to be present in the build.  Uses should not be able to map any other function to the pin used by pid audio.

The actual code size difference for enabling PID audio is quite small.

configs/default/SPRO-SPRACINGH7EXTREME.config
blckmn: USE_PINIO is an example of such an option / gate. The user can choose now - with cloud build - to have it included in the firmware or not.

Essentially users now have both compile time and runtime options, whereas before they only had runtime.

hydra: the problem is users don't know what hardware engineers and target authors intended and or how the circuits work.  it's often not OK to just leave circuits at some random default or the pin be re-assigned accidentally to another function by a user, hence the requirement for a base target on which users can build upon, but not remove from.

For this specific issue buttons are a hardware feature which requires correct initialization prior to a config to work.

Things like a GPS support, or some legacy RX protocol that no-one needs are perfect examples of what can be compiles out, but not things like `USE_PID_AUDIO` and `USE_BUTTONS`.

blckmn: It's your target so you can force it if you want.

blckmn (approved):

haslinghuis (approved):

configs/default/SPRO-SPRACINGH7EXTREME.config
haslinghuis: @hydra H7RF is a H730 target. Please proceed removing the defines from H750 target.h if applicable. We need to document this for target maintainers.

configs/default/SPRO-SPRACINGH7EXTREME.config
haslinghuis: Also need to be in the same documentation

configs/default/SPRO-SPRACINGH7EXTREME.config
haslinghuis: We brought back defining of baro on target now.

PR #762, opened 2022-12-29
added IFLIGHT BLITZ X1 target

Guidus93:
added IFLIGHT BLITZ X1 target


configs/default/IFRC-IFLIGHT_BLITZ_F722_X1.config
```diff
@@ -0,0 +1,123 @@
...
+feature OSD
+
+# serial
+serial 1 64 115200 57600 0 115200
+
+# master
+set mag_bustype = I2C
+set mag_i2c_device = 2
+set baro_bustype = I2C
+set baro_i2c_device = 1
```
haslinghuis, line 107: If this board has a baro please also #define after GYRO/ACC defines

haslinghuis (approved):

blckmn (approved):

PR #792, opened 2023-01-04
Configuration updates for cloud build, add ELRS target

MJ666:


configs/default/AFNG-ALIENFLIGHTNGF7.config
```diff
@@ -1,8 +1,13 @@
-# Betaflight / STM32F7X2 (S7X2) 4.1.0 Oct 16 2019 / 11:58:45 (c37a7c91a) MSP API: 1.42
+# Betaflight / STM32F7X2 (S7X2) 4.3.2 Nov 28 2022 / 07:30:19 (60c9521) MSP API: 1.44
 
 #define USE_GYRO_SPI_MPU6500
 #define USE_ACC_SPI_MPU6500
+#define USE_GYRO_SPI_ICM20602
+#define USE_ACC_SPI_ICM20602
```
haslinghuis, line 6: ICM20602 is handled by MPU6500

```suggestion
#define USE_GYRO_SPI_MPU6500
#define USE_ACC_SPI_MPU6500
```

haslinghuis (changes requested):

configs/default/AFNG-ALIENFLIGHTNGF7.config
KarateBrot: Aren't they independent gyros?

MJ666: There are a few different AFNG boards using different hardware but same firmware. Official board designs are published here: https://www.alienflightng.com/
Anyhow there are some more boards out in the wild wich are based on the AFNG design. They are build by AFNG team members and in the AF community. These are boards with MPU6000 or MPU9250. These are unpublished porotypes which likely can be handled by custom defines in the configurator.

haslinghuis (approved):

configs/default/AFNG-ALIENFLIGHTNGF7.config
haslinghuis: @KarateBrot these gyro's are not much used anymore.
If you look in gyro code you see the `MPU6500` handles `ICM2060x` also:

https://github.com/betaflight/betaflight/blob/a3a21d9233660d5f9bcd47d62f3a550e27b04e6f/src/main/sensors/gyro_init.c#L402-L431

KarateBrot: Okay I see. I still think it makes sense to include the correct one for completeness. Just `#define` whatever is on the board and let the pre-compiler do its job afterwards.

haslinghuis: @KarateBrot Do you want me to make changes in firmware to make this happen?

KarateBrot: It already does that in firmware, that's my point. We don't need to change the define to MPU6500.

haslinghuis: https://github.com/betaflight/betaflight/pull/12161

KarateBrot: @MJ666 Thanks to haslinghuis you can now `#define` the actual gyro that is on board (ICM20602). But you can also leave it as MPU6500. The board config will still work. I prefer to have the actual gyro name in the board config but it's your call 👍

blckmn (approved):

configs/default/AFNG-ALIENFLIGHTNGF7.config
MJ666: Will make a new PR to update and add all sensors used on different AFNG boards.

MJ666:
The Mag (SPI AK8963) on AFNG controllers is only working if MAG option is configured during build. 

The internal Mag of the SPI MPU9250 is also not working anymore. There is no difference if Mag is selected or not during cloud build. It is working with BF 4.3.2 and 4,2,11.

PR #794, opened 2023-01-07
MTKS-MATEKF405TE has onboard FLASH, not SD

SteveCEvans:
FLASH was not appearing using cloud build.

### DIFF
--- a/configs/default/MTKS-MATEKF405TE.config
+++ b/configs/default/MTKS-MATEKF405TE.config
@@ -4,7 +4,7 @@
 #define USE_ACC_SPI_ICM42688P
 #define USE_BARO_DPS310
 #define USE_MAX7456
-#define USE_SDCARD
+#define USE_FLASH_M25P16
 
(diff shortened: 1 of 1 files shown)


TheIsotopes:
@SteveCEvans both would be the best

MTK F405-TE using SD 
MTK F405-miniTE using Flash

haslinghuis:
We also need to change set blackbox_device = SPIFLASH to SDCARD

But then it still does not show up for me? Only tested with adding SDCARD using custom defines and changing above setting.

SteveCEvans:
Agreed, I think a new target makes most sense or users of the MINI (such as myself) would see a failure to initialise the SD on the Blackbox tab.

haslinghuis:
Can't confirm SDC works on the BIG one, as this should need `set blackbox_device = SPIFLASH` to `SDCARD`

MINI target is here: https://github.com/betaflight/unified-targets/pull/795

PR #850, opened 2023-01-29
SPRacingH7RF - First cut of a unified target.

hydra:
Requires PR https://github.com/betaflight/betaflight/pull/11825

### DIFF
--- a/configs/default/SPRO-SPRACINGH7RF.config
+++ b/configs/default/SPRO-SPRACINGH7RF.config
@@ -1,0 +1,327 @@
+# Betaflight / STM32H730 (SP7R) 4.4.0 Jan 28 2023 / 21:17:30 (d78f0c9183) MSP API: 1.45
+
+#define TARGET_BOARD_IDENTIFIER "SP7R"
+#define USBD_PRODUCT_STRING "SPRacingH7RF"
+
+#
+# Settings required to boot
+#
+# Note: hardware for all possible config storage subsystems needs to be known and initialised BEFORE loading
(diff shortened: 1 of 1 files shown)


configs/default/SPRO-SPRACINGH7RF.config
```diff
@@ -0,0 +1,321 @@
...
+#define USBD_PRODUCT_STRING "SPRacingH7RF"
+
+#
+# Settings required to boot
+#
+# Note: hardware for all possible config storage subsystems needs to be known and initialised BEFORE loading
+# a config from it can work. (SPI/OCTOSPI/SDCARD).
+#
+
+#define EEPROM_SIZE=8192
```
haslinghuis, line 13: ISO C99 requires whitespace after the macro name

```suggestion
#define EEPROM_SIZE = 8192
```

haslinghuis (changes requested):

configs/default/SPRO-SPRACINGH7RF.config
hydra: yeah, I had noticed your change to the H7EXTREME target and made the same change here, but due to a git snafu the commit didn't make it to the PR branch.

In this case there's no need for the `=` at all and the corresponding change should also be set it in H7EXTREME target.

i.e.

`#define EEPROM_SIZE 8192`

hydra: @haslinghuis See also https://github.com/betaflight/unified-targets/pull/857

haslinghuis (approved):

KarateBrot (approved):

PR #902, opened 2023-02-16
Add support for a range of sensors for NERC and GEPR targets

Linjieqiang:


haslinghuis:
Think it is not the intention to apply defines that are the same for all targets, otherwise it would be better to do this centrally as we already have put a lot of effort to filter the defines for each configuration.

sugaarK:
Agree with  @haslinghuis. The targets should have defines that are used on the fc using the target not all the drivers.. this makes the way out cloud builder works pointless

PR #924, opened 2023-02-28
Modify MatekF405TE series targets

MATEKSYS:
Previous MatekF405TE and MatekF405TEMINI both don't have SD card definitions.
Rename MatekF405TEMINI with MatekF405TE_SD,  unify the naming with INAV MatekF405TE & MatekF405TE_SD targets

### DIFF
--- a/configs/default/MTKS-MATEKF405TE.config
+++ b/configs/default/MTKS-MATEKF405TE.config
@@ -4,7 +4,7 @@
 #define USE_ACC_SPI_ICM42688P
 #define USE_BARO_DPS310
 #define USE_MAX7456
-#define USE_SDCARD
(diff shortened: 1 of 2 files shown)


haslinghuis (approved):

sugaarK (approved):

sugaarK:
@MATEKSYS 

can we get some of the f4v2 and h7 wing boards? we are releasing full wing code in 4.6 and would be good to have some more targets other than the speedybee ones. also could we get some of the digital airspeed sensors as well  as i'd like to add them as well, we are going to look at adding CAN sensors so any thing that helps with that would be great.  as well as wings we will be looking at way point nav next release as well 

email me at sugarK@betaflight.com

PR #925, opened 2023-02-28
CLRACINGF4 add W25Q128FV

klutvott123:
for the bardwell F4.
Not sure if we should make a clone config or not for this as I believe the FC is designed by the same person that designed the clracing FC so it's not really a clone.
Maybe it should have it's own config, but I'm not sure who's the manufacturer here. Is it JB? RDQ? clracing? lol

### DIFF
--- a/configs/default/CLRA-CLRACINGF4.config
+++ b/configs/default/CLRA-CLRACINGF4.config
@@ -6,6 +6,7 @@
 #define USE_ACC_SPI_MPU6000
 #define USE_MAX7456
 #define USE_SDCARD
+#define USE_FLASH_W25Q128FV
 
 board_name CLRACINGF4
 manufacturer_id CLRA


haslinghuis:
We could call it `CLRACINGF4_SD`

klutvott123:
then CLRACINGF4_SD would be the target for the actual clracing f4, and the bardwell f4 would continue to use CLRACINGF4. lol. doesn't seem right to me.

blckmn (approved):

blckmn:
> then CLRACINGF4_SD would be the target for the actual clracing f4, and the bardwell f4 would continue to use CLRACINGF4. lol. doesn't seem right to me.

What is the bardwell F4 being marketed as? I'd be more inclined to leave CLRACINGF4 as is, and created the bardwell F4 closer to what it is being marketed as.

haslinghuis (approved):

haslinghuis:
@joshuabardwell

PR #931, opened 2023-03-05
fixed the wrong spi bus for elrs spi rx

freasy:
Fixes the wrong SPI BUS config setting

### DIFF
--- a/configs/default/HGLR-HGLRCF411SX1280.config
+++ b/configs/default/HGLR-HGLRCF411SX1280.config
@@ -86,6 +86,7 @@ dma pin B07 0
 
 # feature
 feature RX_SPI
+feature -RX_SERIAL
 feature TELEMETRY
 feature OSD
 
--- a/configs/default/HGLR-HGLRCF411SX1280.config
+++ b/configs/default/HGLR-HGLRCF411SX1280.config
@@ -90,7 +90,7 @@ feature TELEMETRY
 feature OSD
 
 # master
-set rx_spi_bus = 1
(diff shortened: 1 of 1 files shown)


haslinghuis (approved):

configs/default/HGLR-HGLRCF411SX1280.config
```diff
@@ -86,6 +86,7 @@ dma pin B07 0
 
 # feature
 feature RX_SPI
+feature -RX_SERIAL
```
haslinghuis, line 89: We can leave this out as we already use feature RX_SPI

freasy: Serial Rx is set as default in the new build system for 4.4. This is currently needed to default to SPI RX. Otherwise serial rx is preselected after applying custom defaults.

haslinghuis (approved):

configs/default/HGLR-HGLRCF411SX1280.config
haslinghuis: You also can define:

    #define DEFAULT_RX_FEATURE              FEATURE_RX_SPI
    #define RX_SPI_PROTOCOL                 EXPRESSLRS

freasy: Yes i know, but i think since it's all overhauled for 4.5 and i will make a PR for this board there too, this is enough for now.

KarateBrot (approved):

PR #937, opened 2023-03-07
Add new Foxeer F722 V4 MPU6000 Target

Johnsnow-foxeer:
Add new target base on MPU6000 gyro, barometer is on board.

### DIFF
--- a/configs/default/FOXE-FOXEERF7V4M6000.config
+++ b/configs/default/FOXE-FOXEERF7V4M6000.config
@@ -1,0 +1,116 @@
+# Betaflight / STM32F7X2 (S7X2) 4.0.0 Apr  4 2019 / 00:54:12 (e44f75b4a) MSP API: 1.41
+
+#define USE_GYRO_SPI_MPU6000
+#define USE_ACC_SPI_MPU6000
+#define USE_BARO_DPS310
+#define USE_FLASH_W25Q128FV
+#define USE_MAX7456
+
+board_name FOXEERF7V4M6000
+manufacturer_id FOXE
+
+# resources
+resource BEEPER 1 A04
+resource MOTOR 1 A09
+resource MOTOR 2 A08
+resource MOTOR 3 C09
+resource MOTOR 4 C08
+resource PPM 1 B07
+resource LED_STRIP 1 A15
+resource SERIAL_TX 1 B06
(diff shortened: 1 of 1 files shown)


configs/default/FOXE-FOXEERF7V4M6000.config
```diff
@@ -0,0 +1,116 @@
+# Betaflight / STM32F7X2 (S7X2) 4.0.0 Apr  4 2019 / 00:54:12 (e44f75b4a) MSP API: 1.41
+
+#define USE_GYRO_SPI_MPU6000
+#define USE_ACC_SPI_MPU6000
+#define USE_BARO_DPS310
+#define USE_FLASH_W25Q128FV
+#define USE_MAX7456
+
+board_name FOXEERF7V4M6000
```
haslinghuis, line 9: Should we just call it V5 as this looks a lot like V4 + BARO
Or we could add the BARO to V4

Johnsnow-foxeer: V4, please, because the board is printed V4 already

haslinghuis: Okay NP... just add the defines to the V4. No need for a new target :) - will update for 4.5 too.

Johnsnow-foxeer: Added the defines, tested, the barometer doesn't work, so add a new target

haslinghuis: Nah, we are fixing that.

sugaarK (approved):

haslinghuis:
@sugaarK we will be updating V4 with the MPU6000 instead of creating a new one as it's already in production as V4.

configs/default/FOXE-FOXEERF7V4M6000.config
haslinghuis: @Johnsnow-foxeer be sure to enable expert mode on the firmware flasher tab (it's hidden beneath show release candidates. Then add in custom defines: `ACC_SPI_MPU6000 GYRO_SPI_MPU6000` and flash 4.4.1 (or 4.5)

sugaarK:
I did that already I'll close this

PR #945, opened 2023-03-11
Added USE_CAMERA_CONTROL to all foxeer boards

limonspb:
we have half of the all boards with `resource CAMERA_CONTROL`
but only a couple actually have 
`#define USE_CAMERA_CONTROL`
this causes error in CLI when resetting to defaults

This is a fix for just foxeer boards

@haslinghuis  @blckmn


KarateBrot:
Do the Foxeer boards even use camera control?

sugaarK (approved):

PR #949, opened 2023-03-14
add BMP280 to SKYSTARSH7HD

nerdCopter:
* as per https://build.betaflight.com/api/support/4b87b4a5-09bc-4823-8e0c-3aebd3c164c5


haslinghuis (approved):

KarateBrot (approved):

Some-Won-1:
Baro NOT WORKING with BF 4.4. In the configurator "use barometer" rejects a lock every time even with basic defaults.  I have verified this to be the case on two new stacks. Also, Power to RX 7 (s.port receiver) and GPS does not energize with USB until  a battery is plugged in.  I'm almost ready to RMA these units!

PR #950, opened 2023-03-16
Added SUB250F411 target

Guidus93:
I added SUB250F411 target, as it's a controller for BNF DSHOT it's already enabled and there are some settings (such as idle, antigravity and UARTs already in the config file).

### DIFF
--- a/configs/default/SU25-SUB250F411.config
+++ b/configs/default/SU25-SUB250F411.config
@@ -1,0 +1,108 @@
+# Betaflight / STM32F411 (S411) 4.3.2 Nov 28 2022 / 07:27:10 (60c9521) MSP API: 1.44
+
+#define USE_ACCGYRO_BMI270
+#define USE_GYRO_SPI_ICM42688P
+#define USE_ACC_SPI_ICM42688P
+#define USE_BARO_DPS310
+#define USE_BARO_BMP280
+#define USE_FLASH_W25Q128FV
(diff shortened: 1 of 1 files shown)


configs/default/SU25-SUB250F411.config
```diff
@@ -0,0 +1,108 @@
...
+set flash_spi_bus = 2
+set gyro_1_bustype = SPI
+set gyro_1_spibus = 1
+set gyro_1_sensor_align = CW270
+set gyro_1_align_yaw = 2700
+
+profile 0
+
+# profile 0
+set anti_gravity_gain = 0
```
sugaarK, line 108: why would you set anti grav to zero????

sugaarK (changes requested):

configs/default/SU25-SUB250F411.config
Guidus93: That’s what the manufacturer asked me to do. It’s probably a BNF tiny quad

sugaarK: Still doesn’t make any sense if they have a preset tune then that’s where it should be. And frankly as a good tuner I don’t see what they are upto

Guidus93: Removed the antigravity setting

KarateBrot (approved):

PR #1017, opened 2023-05-25
fix 3 hakrc board definitions

freasy:
Fix 2 boards per request.

Fix F722V2 board, that will freeze after 4.4.1

gyro=both seems broken in 4.4.1.


blckmn (approved):

configs/default/HARC-HAKRCF722V2.config
```diff
@@ -115,7 +115,8 @@ feature OSD
 serial 5 64 115200 57600 0 115200
 
 # master
-set gyro_to_use = BOTH
+# this does not work in 4.4.1
+# set gyro_to_use = BOTH
```
haslinghuis, line 119: This will break the target for 4.4.0

haslinghuis (changes requested):

configs/default/HARC-HAKRCF722V2.config
blckmn: Will it? It'll only disable 1 gyro.

haslinghuis: Cannot confirm as I don't have any dual gyro board for testing.

haslinghuis (approved):

PR #1019, opened 2023-05-26
Create DARWINF722HD_SD target

DMAXYANG:
It changed the flash to sd card on the DARWINF722HD target and added the 42688P gyro, 
Thank you

### DIFF
--- a/configs/default/DAFP-DARWINF722HD_SD.config
+++ b/configs/default/DAFP-DARWINF722HD_SD.config
@@ -135,15 +135,12 @@ feature OSD
 aux 0 40 2 1300 2100 0 0
 
 # master
-set dyn_notch_count = 1
-set dyn_notch_q = 500
 set mag_bustype = I2C
 set mag_i2c_device = 1
 set mag_hardware = NONE
 set baro_bustype = I2C
 set baro_i2c_device = 1
 set baro_hardware = NONE
-set serialrx_provider = CRSF
 set blackbox_sample_rate = 1/16
 set blackbox_device = SDCARD
 set dshot_burst = ON
@@ -156,7 +153,6 @@ set beeper_inversion = ON
 set beeper_od = OFF
 set sdcard_mode = SPI
 set sdcard_spi_bus = 3
-set debug_mode = GYRO_SCALED
 set max7456_spi_bus = 2
 set pinio_box = 40,41,255,255
 set flash_spi_bus = 3
--- a/configs/default/DAFP-DARWINF722HD_SD.config
+++ b/configs/default/DAFP-DARWINF722HD_SD.config
@@ -1,0 +1,166 @@
+# Betaflight / STM32F7X2 (S7X2) 4.3.2 Nov 28 2022 / 07:30:19 (60c9521) MSP API: 1.44
+
+#define USE_ACC
+#define USE_ACC_SPI_MPU6500
+#define USE_ACC_SPI_MPU6000
+#define USE_ACC_SPI_ICM42688P
+#define USE_GYRO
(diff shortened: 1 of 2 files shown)


DMAXYANG:
> * the name contains `HD_SD`, is this wanted?
> * if nothing else changed except the gyro and sdcard, then why not just add those new things to the existing definitions?
Because some Custom Settings were also modified, these new Settings would affect pilots using the old target board

haslinghuis:
We need a new target as changing from flash to sdcard also requires default blackbox device change but would suggest renaming it to DARWINF722HDV2.

DMAXYANG:
> We need a new target as changing from flash to sdcard also requires default blackbox device change but would suggest renaming it to DARWINF722HDV2.

Hello,
I mainly consider that the V1 version will not be discontinued, and its Layout is also very different from the V1 version


Thanks

configs/default/DAFP-DARWINF722HD_SD.config
```diff
@@ -0,0 +1,166 @@
...
+feature TELEMETRY
+feature LED_STRIP
+feature OSD
+
+# aux
+aux 0 40 2 1300 2100 0 0
+
+# master
+set dyn_notch_count = 1
+set dyn_notch_q = 500
```
haslinghuis, line 139: Remove tuning (we have presets for tuning)
```suggestion


```

configs/default/DAFP-DARWINF722HD_SD.config
```diff
@@ -0,0 +1,166 @@
...
+set mag_bustype = I2C
+set mag_i2c_device = 1
+set mag_hardware = NONE
+set baro_bustype = I2C
+set baro_i2c_device = 1
+set baro_hardware = NONE
+set serialrx_provider = CRSF
```
haslinghuis, line 146: This will be defined using cloud build
```suggestion

```

configs/default/DAFP-DARWINF722HD_SD.config
```diff
@@ -0,0 +1,166 @@
...
+set dshot_bidir = ON
+set motor_pwm_protocol = DSHOT600
+set current_meter = ADC
+set battery_meter = ADC
+set ibata_scale = 125
+set beeper_inversion = ON
+set beeper_od = OFF
+set sdcard_mode = SPI
+set sdcard_spi_bus = 3
+set debug_mode = GYRO_SCALED
```
haslinghuis, line 159: Remove debug
```suggestion

```

configs/default/DAFP-DARWINF722HD_SD.config
```diff
@@ -0,0 +1,166 @@
...
+set max7456_spi_bus = 2
+set pinio_box = 40,41,255,255
+set flash_spi_bus = 3
+set gyro_1_bustype = SPI
+set gyro_1_spibus = 1
+set gyro_1_sensor_align = CW180
+set gyro_1_align_yaw = 1800
```
haslinghuis, line 166: Add new line
```suggestion
set gyro_1_align_yaw = 1800

```

haslinghuis (changes requested):

haslinghuis (approved):

haslinghuis:
Agree

DMAXYANG:
> i still prefer `DARWINF722HDV2` rather than `DARWINF722HD_SD` only because pilots will read "SD" as standard definition not "sdcard". but this is just an opinion. thoughts @haslinghuis ?

All right, it's just a naming convention.

sugaarK (approved):

haslinghuis (approved):

PR #1030, opened 2023-06-01
Update board Aocodarc f722mini

aocodarc:
1. Add s8 motor and modify the motor sequence
2. Define the MPU6000 and DPS310
3. The default light strip is set


haslinghuis:
See: https://github.com/betaflight/unified-targets/pull/1031

PR #1040, opened 2023-06-09
set gyro orientation

x4FF3:
Hi

after contacting GEPRC, this is the desired default orientation.

KR
David

### DIFF
--- a/configs/default/GEPR-TAKERG4AIO.config
+++ b/configs/default/GEPR-TAKERG4AIO.config
@@ -91,7 +91,6 @@ feature AIRMODE
 
 
 # master
-set baro_i2c_device = 1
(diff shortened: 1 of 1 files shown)


configs/default/GEPR-TAKERG4AIO.config
```diff
@@ -91,7 +91,6 @@ feature AIRMODE
 
 
 # master
-set baro_i2c_device = 1
```
haslinghuis, line 94: Why remove baro i2c device setting?

x4FF3: removed baro drivers, as no baro is present on final version.

haslinghuis (approved):

PR #1070, opened 2023-07-20
Add config file: SPBE-SPEEDYBEEF405MINI.config

SEA729:
We have successfully verified and thoroughly tested the config file "SPEEDYBEEF405MINI." Kindly consider approving it. We really appreciate your help.


sugaarK:
@SEA729 you are from speedybee? We have some hardware issues with one of your products that is not software based.. we’d like to get it sorted as it’s consuming our time with support for it..

SEA729:
@sugaarK Yes, I'm Sea from SpeedyBee. Sorry for the unconvenience that caused by our product:-( Which flight controller did you mean? And what was the problem? May I have your Discord account so we can't communicate more easily?

haslinghuis (changes requested):
This needs a new PR (as it's based on master - not a feature branch)

image

PR #1074, opened 2023-07-22
Update DMA stream on SPEEDYBEEF7V3

haslinghuis:
Fixes: https://github.com/betaflight/betaflight/issues/12965

### DIFF
--- a/configs/default/SPBE-SPEEDYBEEF7V3.config
+++ b/configs/default/SPBE-SPEEDYBEEF7V3.config
@@ -87,7 +87,7 @@ dma SPI_MOSI 3 0
 # SPI_MOSI 3: DMA1 Stream 5 Channel 0
 dma SPI_TX 3 0
 # SPI_TX 3: DMA1 Stream 5 Channel 0
-dma ADC 1 0
+dma ADC 1 1
 # ADC 1: DMA2 Stream 0 Channel 0
 dma pin A15 0
 # pin A15: DMA1 Stream 5 Channel 3


nerdCopter (approved):

Martivip (approved):

blckmn (approved):

Martivip:
Hi ok i see that was merged to master and after clean flash to 4.4.2 an applying custom defaults still under `dump` I receiving `dma ADC 1 0` not `dmaADC 1 1`

haslinghuis:
Probably the build is cached and not picking up changes. Try adding Acro_Trainer from Other Options to the build for example.

Martivip:
> Probably the build is cached and not picking up changes. Try adding Acro_Trainer from Other Options to the build for example.

@haslinghuis Thank You it helps is ok, do you know how to clear that cached builds in the configurator on Mac ? I'm asking because configurator reinstallation with deleting Configurator from Application Support folder don't clear that cached builds

haslinghuis:
Sorry build is cached -> in the cloud.

Martivip:
> Sorry build is cached -> in the cloud.

OK but do you know any PR is addressed to resolve that issues ? because right now in case of unified_targets updates those updates are not pushed to the builds

Martivip:
@haslinghuis or maybe in case on unified-target update admins of the cloud should remove all builds for this specific target  then users will start to caching the new ones with updated target, what do you think about that ?

haslinghuis:
@blckmn can we invalidate cached builds in case of configuration updates?

blckmn:
> @blckmn can we invalidate cached builds in case of configuration updates?

In terms of these unified configs, there's no need.

For the config.h it's already happening.

PR #1079, opened 2023-07-24
Add a new target TUNERCF411

TUNERC-Aria:
Add a new target TUNERC_F411


haslinghuis (approved):

nerdCopter (approved):

haslinghuis (changes requested):
Working on adding the config file. Should we rename this to TUNERCF411 to be in line with TUNERCF405 target?

Second we need to remove:
image

haslinghuis (approved):

steronius (approved):

nerdCopter (approved):

PR #1081, opened 2023-07-28
fix spi mosi, miso

freasy:
old SPI definitions in custom defines fixed


KarateBrot:
SDI and SDO are the new definitions. But I guess they silently snuck into the old unified-targets

KarateBrot (approved):

nerdCopter (approved):

PR #1084, opened 2023-08-03
Only few with #wiki string left over

HThuren:
remove #wiki in most files, but HBRO-KAKUTEF4.config, HBRO-KAKUTEF7.config and HBRO-KAKUTEH7.config
Link in files are https://betaflight.com/docs/wiki/boards/KAKUTExx

### DIFF
--- a/configs/default/HBRO-KAKUTEF7.config
+++ b/configs/default/HBRO-KAKUTEF7.config
@@ -1,7 +1,7 @@
 # Betaflight / STM32F745 (S745) 4.0.0 Apr  3 2019 / 14:32:23 (22b9f3453) MSP API: 1.41
 
 #mcu STM32F745
-#wiki https://betaflight.com/docs/boards/KAKUTEF7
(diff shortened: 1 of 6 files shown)


configs/default/HBRO-KAKUTEF7.config
```diff
@@ -1,7 +1,7 @@
 # Betaflight / STM32F745 (S745) 4.0.0 Apr  3 2019 / 14:32:23 (22b9f3453) MSP API: 1.41
 
 #mcu STM32F745
-#wiki https://betaflight.com/docs/boards/KAKUTEF7
+#wiki https://betaflight.com/docs/wiki/boards/KAKUTEF7
```
haslinghuis, line 4: Without this line we also reach board documentation. This can be removed like all other #wiki directives.

HThuren: Hmm, something wnet wrong in my PR :-(
Just keep those for reference :-)

haslinghuis (approved):

nerdCopter (approved):

PR #1103, opened 2023-08-14
add HGLR-HGLRC_F722_AIO_40V1.config

HGLRC-T:
Create HGLR-HGLRC_F722_AIO_40V1 (include hw define)

### DIFF
--- a/configs/default/HGLR-HGLRC_F722_AIO_40V1.config
+++ b/configs/default/HGLR-HGLRC_F722_AIO_40V1.config
@@ -1,0 +1,144 @@
+# Betaflight / STM32F7X2 (S7X2) 4.4.0 Feb 15 2023 / 06:10:00 (4605309d8) MSP API: 1.45
+
+#define USE_ACC
+#define USE_ACC_SPI_MPU6000
+#define USE_ACC_SPI_ICM42688P
+#define USE_ACCGYRO_BMI270
+#define USE_BARO
+#define USE_BARO_BMP280
+#define USE_BARO_DPS310
+#define USE_FLASH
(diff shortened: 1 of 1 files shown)


configs/default/HGLR-HGLRC_F722_AIO_40V1.config
```diff
@@ -0,0 +1,144 @@
...
+#define USE_BARO_BMP280
+#define USE_BARO_DPS310
+#define USE_FLASH
+#define USE_FLASH_W25Q128FV
+#define USE_GYRO
+#define USE_GYRO_SPI_MPU6000
+#define USE_GYRO_SPI_ICM42688P
+#define USE_MAX7456
+
+board_name HGLRC_F722_AIO_40V1
```
haslinghuis, line 17:
```suggestion
board_name HGLRCF722AIO
```

nerdCopter:
@HGLRC-T , all PR's must be made from branches, not from `master`

configs/default/HGLR-HGLRC_F722_AIO_40V1.config
HGLRC-T: > 

We will have two FC models with different functions in the future, and we hope to maintain this name to avoid confusion about the target

haslinghuis:
Also naming should be HGLRCF722AIO

HGLRC-T:
> Also naming should be HGLRCF722AIO

We will have two FC models with different functions in the future, and we hope to maintain this name to avoid confusion about the target

haslinghuis:
Second model can use HGLRCF722AIOV2

PR #1114, opened 2023-08-15
add new target FLMO-STELLARF7.config

cvetaevvitaliy:
add new target FLMO-STELLARF7.config

### DIFF
--- a/configs/default/FLMO-STELLARF7.config
+++ b/configs/default/FLMO-STELLARF7.config
@@ -1,0 +1,134 @@
+# Betaflight / STM32F7X2 (S7X2) 4.4.2 Aug 14 2023 / 20:54:15 (23d066d08) MSP API: 1.45
+
+#define USE_GYRO
+#define USE_GYRO_SPI_MPU6000
+#define USE_ACC
+#define USE_ACC_SPI_MPU6000
+#define USE_ACC_SPI_ICM20602
+#define USE_GYRO_SPI_ICM20602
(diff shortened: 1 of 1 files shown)


configs/default/FLMO-STELLARF7.config
```diff
@@ -0,0 +1,134 @@
...
+dma pin B10 0
+# pin B10: DMA1 Stream 1 Channel 3
+dma pin B11 0
+# pin B11: DMA1 Stream 7 Channel 3
+
+# feature
+feature OSD
+feature RX_SERIAL
+feature TELEMETRY
+feature GPS
```
haslinghuis, line 97: Remove cloud build option

```suggestion

```

cvetaevvitaliy: I need to read the documentation, what is now in cloud build and what is in the config file 😁

haslinghuis (approved):

nerdCopter:
@haslinghuis ... i've not memorized all the recommendations...

are the serial  1, 2, 4 okay to define here?
beacon's?
dshot?
osd_displayport_device?

i dont recall what is firmware-default-set and what is required to be defined.

haslinghuis:
@nerdCopter Other unified targets use these settings too.
For BNF/PNP it's convenient and cloud build would override.
We can conduct an all-purpose cleanup when the settings become a problem

nerdCopter (approved):

PR #1144, opened 2023-09-07
Add F7miniV2 file

aocodarc:
Add F7miniV2 file


haslinghuis:
Please don't open new PR's for the same work.
There are many defaults which needs to be removed and the file has illegal DOS line endings.

configs/default/SJET-AOCODAF722MINIV2.config
```diff
@@ -0,0 +1,162 @@
...
+
+
+# master
+set motor_pwm_protocol = DSHOT600
+set serialrx_provider = SBUS
+set gyro_to_use = FIRST
+set align_mag = DEFAULT
+set mag_align_roll = 0
+set mag_align_pitch = 0
+set mag_align_yaw = 0
```
haslinghuis, line 119: Please don't open new PR's for the same work.

There are many defaults which needs to be removed.
The file has illegal DOS line endings.

haslinghuis (changes requested):

configs/default/SJET-AOCODAF722MINIV2.config
aocodarc: I have noticed this problem, thank you for reminding me

nerdCopter: @aocodarc , there are may ways to fix branches/PR'
please look into `git reset [--soft|--hard]` and/or rebasing and/or squashing.
once fixed you can force push (`git push --force`).


also this may be helpful for your overall git environment:
```
git config --global core.autocrlf true    ## use "input" for Linux or use "true" for Windows
git config --global core.safecrlf warn
git config --global core.whitespace cr-at-eol
git config --global core.filemode false
git config --global help.autocorrect true
```

PR #1171, opened 2023-10-26
GEPRCF722_BT_HD: Fix the sensor_align to 0

Linjieqiang:
When this board was developed, the BMI270 gyroscope was used. The current mass production gyroscope is ICM-42688, so the inconsistency in the direction of the machine head caused by updating the firmware to the user. The purpose of the modification is to allow the user to maintain the default direction of the firmware after updating the firmware. The FC’s machine heads are in the same direction, reducing user confusion.


nerdCopter (dismissed):

configs/default/GEPR-GEPRCF722_BT_HD.config
```diff
@@ -155,6 +155,6 @@ set pinio_config = 129,1,1,1
 set pinio_box = 0,40,41,42
 set gyro_1_bustype = SPI
 set gyro_1_spibus = 1
-set gyro_1_sensor_align = CW270
-set gyro_1_align_yaw = 2700
+set gyro_1_sensor_align = CW0
+set gyro_1_align_yaw = 0
```
haslinghuis, line 159: This is default

```suggestion
```

haslinghuis (approved):

nerdCopter (approved):

Issue #1149, opened 2023-09-17
FreeAxis 25A no Gyro, no baro, no blackbox target MatekF411


IKKI-FPV:
*** Dati per il supporto inviati *** 
Id: "793e2e7e-d244-4b4c-9f3b-4c2599f57804"

https://it.aliexpress.com/item/1005005844783456.html

nerdCopter:
two reports on that link about failing after update.  manufacturer may have created custom definition/hex.

https://build.betaflight.com/api/support/793e2e7e-d244-4b4c-9f3b-4c2599f57804 indeed show no gyro suing `MATEKF411`

https://build.betaflight.com/api/builds/a95d31bd5684a9366c1e0a7cf123ccb9/log shows all the gyro and blackbox chips compiled. 

MATEKF411 defines only the mpu6000 by default: https://github.com/betaflight/config/blob/master/configs/MATEKF411/config.h

do you have a `dump` from the original firmware saved?

IKKI-FPV:
Hello and thanks for the reply.  Unfortunately no, I stupidly didn't save the dump before updating.  I'm asking the seller for a copy with little success

PR #1173, opened 2023-10-26
Create FRXS-FREEAXISF4.config

ot0tot:
Adds target for FreeAxis F4 25A AIO FC based on MATEKF411 target with the addition of a flash chip and BMI270.

Needs to be tested, gyro orientation verified, and FreeAxis vendor/board name approved.

Solves: betaflight/unified-targets#1149

Discord issue: https://discord.com/channels/868013470023548938/1160296288609390592

Thanks to J. R. @sixtyfive for the factory config https://discord.com/channels/868013470023548938/868013470518505494/1167219861601124392


sixtyfive:
If/once there's a build somewhere, will happily confirm gyro orientation and test fly.

sixtyfive:
Once there's a build somewhere, will happily confirm gyro orientation and test fly.

nerdCopter (approved):

nerdCopter:
redacting my approval until manufacturer initials are decided upon by a core-dev.  I believe this manufacturer is a "Cloner".

configs/default/FRXS-FREEAXISF4.config
```diff
@@ -0,0 +1,122 @@
...
+
+#define USE_GYRO
+#define USE_ACC
+#define USE_ACCGYRO_BMI270
+#define USE_FLASH
+#define USE_FLASH_W25N01G
+#define USE_MAX7456
+
+board_name FREEAXISF4
+manufacturer_id FRXS
```
haslinghuis, line 13:
```suggestion
board_name MATEKF411_CLONE
manufacturer_id CLNE
```
Before submitting new targets we need a manufacturer. As this board again uses another manufacturers work, rather I'd suggest to make a CLNE-MATEKF411 configuration. Then we also can add ICM20602 as used by Eachine Tyro79 (see https://github.com/betaflight/unified-targets/issues/1119#issuecomment-1683989184)

haslinghuis (changes requested):

configs/default/FRXS-FREEAXISF4.config
ot0tot: I believe this board differs from the MATEKF411 target with the addition of flash on SPI2, as well as different timer assignments. Not sure if it would work for the Tyro79 FC.

sixtyfive: Fwiw, physically it looks like no Marek board I was able to find googling around. And won't people search under "F" since F...reeAxis is what they bought it as?

haslinghuis: Cannot spot the timer difference. This PR only changes ACC and GYRO defines compared to the original MATEKF411 config and adds FLASH driver and setting: `flash_spi_bus = 2` which would be shared with OSD. To make it worse the flash chip is using W25N01G.

@SteveCEvans what do you think?

haslinghuis: @ot0tot please only change `manufacturer_id CLNE`

haslinghuis (changes requested):

configs/default/FRXS-FREEAXISF4.config
haslinghuis: Made some additional changes to the config

haslinghuis (approved):

haslinghuis (approved):

nerdCopter (approved):

PR #1205, opened 2023-11-27
Remove JHEF405 target

jhemcu:
In the early days, we were not very good at using git operations, so players uploaded them, but they did not understand the schematics, which resulted in a large number of problems that were inconsistent with the actual circuit boards in the target, which easily caused users to mistakenly flash, so we Hope it can be deleted


nerdCopter:
is `JHEF405PRO` the only target that should be used across all JHEF405 hardware?

haslinghuis (approved):

nerdCopter (approved):

PR #1211, opened 2023-12-04
add baro BMP 280 Support

hakrc1:
add baro BMP 280 Support


haslinghuis:
Please split up in two separate PR's

PR #1213, opened 2023-12-04
add  BMP280 baro Support

hakrc1:
add  BMP280 baro Support


haslinghuis:
PR's based on master are not allowed.

Why open a new PR and then also not create a future branch. Original PR should be reopened and amended.

PR #1245, opened 2024-01-26
Add BETAFPVG473.config

betafpv-engineer:
Add BETAFPVG473 target


haslinghuis:
@ot0tot baro and osd SPI are shared. Please check.

ot0tot:
> @ot0tot baro and osd SPI are shared. Please check.

I haven't looked into the G4 much yet, but I think this will be ok as OSD will primarily use MOSI while baro will primarily use MISO. Do you have a board to test?

haslinghuis:
No samples.

configs/default/BEFH-BETAFPVG473.config
```diff
@@ -0,0 +1,116 @@
...
+#define USE_ACC_SPI_ICM42688P
+#define USE_ACCGYRO_BMI270
+#define USE_BARO
+#define USE_BARO_SPI_BMP280
+#define USE_BARO_SPI_DPS310
+#define USE_FLASH
+#define USE_FLASH_M25P16
+#define USE_MAX7456
+board_name BETAFPVG473
+manufacturer_id BEFH
```
haslinghuis, line 15: Please move to line 3 and separate next block with new line

configs/default/BEFH-BETAFPVG473.config
```diff
@@ -0,0 +1,116 @@
...
+set current_meter = ADC
+set battery_meter = ADC
+set beeper_inversion = ON
+set beeper_od = OFF
+set max7456_spi_bus = 2
+set flash_spi_bus = 3
+set gyro_1_bustype = SPI
+set gyro_1_spibus = 1
+set gyro_1_sensor_align = CW180
+set gyro_1_align_yaw = 1800
```
haslinghuis, line 116: No new line on EOF.

```suggestion
set gyro_1_align_yaw = 1800

```

haslinghuis (changes requested):
Need to create a target in config repo too.

betafpv-engineer:
> No samples.

image
It has been tested that the baro can share an SPI with the OSD.

sugaarK:
> > No samples.
> 
> image It has been tested that the baro can share an SPI with the OSD.

it leads to very poor OSD performance that because the OSD task is so system intensive, can cause issues with overall flight controller performance. we absolutely do not recommend sharing the bus with the OSD chip given its a critical system function!

betafpv-engineer:
> > > No samples.
> > 
> > 
> > image It has been tested that the baro can share an SPI with the OSD.
> 
> it leads to very poor OSD performance that because the OSD task is so system intensive, can cause issues with overall flight controller performance. we absolutely do not recommend sharing the bus with the OSD chip given its a critical system function!

Is it OK for Baro and Flash to share the same SPI?

sugaarK:
> > > > No samples.
> > > 
> > > 
> > > image It has been tested that the baro can share an SPI with the OSD.
> > 
> > 
> > it leads to very poor OSD performance that because the OSD task is so system intensive, can cause issues with overall flight controller performance. we absolutely do not recommend sharing the bus with the OSD chip given its a critical system function!
> 
> Is it OK for Baro and Flash to share the same SPI?

@SteveCEvans  ? your thoughts

betafpv-engineer:
> > > > > No samples.
> > > > 
> > > > 
> > > > image It has been tested that the baro can share an SPI with the OSD.
> > > 
> > > 
> > > it leads to very poor OSD performance that because the OSD task is so system intensive, can cause issues with overall flight controller performance. we absolutely do not recommend sharing the bus with the OSD chip given its a critical system function!
> > 
> > 
> > Is it OK for Baro and Flash to share the same SPI?
> 
> @SteveCEvans ? your thoughts

It has been modified as requested , Baro changed to I2C.

haslinghuis (approved):

nerdCopter (approved):

PR #1246, opened 2024-01-27
Fix AOCODARC_F405V3 target naming convention

lida2003:


haslinghuis (changes requested):
image

Suggest we use AOCODARCF405V3

haslinghuis (approved):

lida2003:
> Suggest we use AOCODARCF405V3

Thanks. Now we have the naming rules for Aocoda-RC products.

nerdCopter (approved):

PR #1258, opened 2024-02-22
Add JHEF722HD board config

jhemcu:


configs/default/JHEF-JHEF722HD.config
```diff
@@ -0,0 +1,137 @@
...
+set beeper_od = OFF
+set max7456_spi_bus = 2
+set pinio_config = 129,129,1,1
+set pinio_box = 40,41,255,255
+set flash_spi_bus = 3
+set gyro_1_bustype = SPI
+set gyro_1_spibus = 1
+set gyro_1_sensor_align = CW270
+set gyro_1_align_yaw = 2700
+set gyro_2_spibus = 1
```
haslinghuis, line 135: SPI sharing is not recommended and can lead to dead locks.

haslinghuis (changes requested):

ctzsnooze (approved):

ctzsnooze:
Approved though I do not have the board to check.  Should not merge until tested to ensure everything works as expected.

configs/default/JHEF-JHEF722HD.config
jhemcu: Thank you for your suggestion. We will avoid this problem in the next board design, but now our products have been launched to the market, and there are no major problems in current use, so I think we can apply for merger.

haslinghuis:
How? Don't have samples to work on?

configs/default/JHEF-JHEF722HD.config
```diff
@@ -0,0 +1,137 @@
...
+dma pin B04 0
+# pin B04: DMA1 Stream 4 Channel 5
+dma pin B03 0
+# pin B03: DMA1 Stream 6 Channel 3
+
+# feature
+feature OSD
+
+# master
+set gyro_to_use = BOTH
```
haslinghuis, line 112: Is this board using both gyro's at once - as this is not reflected in the 4.5 config.

jhemcu: Yes, two gyro sensors are used at the same time. 

image
At the same time, I have also added it to the config file of v4.5

jhemcu:
> How? Don't have samples to work on?

Do you need samples? I can send them to you.

jhemcu:
@haslinghuis  Hi, I've been trying to privately chat with you on discord, but I didn't get your reply, so I can only leave you another message here. This time I will send the design drawing of the board. The board is already being manufactured and is expected to be on the market within this month.


e6b6bad05f35eeea177a1cbd5eb537f
8010fac748746f6d1f52ce8e700babb

haslinghuis (approved):

PR #1296, opened 2024-04-04
Update CODEOWNERS

haslinghuis:


nerdCopter:
- i'm probably not a good choice to approve or disapprove so i would leave such to @blckmn or @sugaarK 
- however, i do see this:
image

haslinghuis:
These errors were there for a long time. Saw this was missing as the targets do not show on release. Not sure if this would make the targets available anyway...

sugaarK (approved):

PR #1318, opened 2024-05-14
Update BEFH-BETAFPVG473

liyouluo:
Hi, please review the target update.

Thank you.

### DIFF
--- a/configs/default/BEFH-BETAFPVG473.config
+++ b/configs/default/BEFH-BETAFPVG473.config
@@ -45,7 +45,6 @@ resource SPI_MOSI 2 B15
 resource SPI_MOSI 3 B05
 resource ADC_BATT 1 A00
 resource ADC_CURR 1 A01
-resource BARO_CS 1 C06
 resource FLASH_CS 1 B09
 resource OSD_CS 1 B12
 resource GYRO_EXTI 1 C04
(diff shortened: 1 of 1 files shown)


sugaarK:
You need to make a config target. 4.5 does not use the unified target branch and the unified farther branch is about to be locked. We have no interest in back porting new hardware to 4.4

nerdCopter (dismissed):

nerdCopter:
documentation: https://betaflight.com/docs/development/manufacturer/creating-configuration
repository: https://github.com/betaflight/config

configs/default/BEFH-BETAFPVG473.config
```diff
@@ -105,7 +117,7 @@ set serialrx_provider = CRSF
 set blackbox_device = SPIFLASH
 set dshot_idle_value = 450
 set dshot_burst = OFF
-set dshot_bitbang = OFF
+set dshot_bitbang = AUTO
```
haslinghuis, line 120: @liyouluo what is the reason for this change? Asking because we had an issue which would be solved in 4.5.1. Have you checked the board with 4.6 development firmware?

liyouluo:
> You need to make a config target. 4.5 does not use the unified target branch and the unified farther branch is about to be locked.

Thanks for the reminder, there is already a BEFH-BETAFPV G473 target in betaflight/config, I will submit a PR to update the target.

liyouluo:
> documentation: https://betaflight.com/docs/development/manufacturer/creating-configuration repository: https://github.com/betaflight/config

OK. I will submit a PR to update the target.

configs/default/BEFH-BETAFPVG473.config
liyouluo: > @liyouluo what is the reason for this change? Asking because we had an issue which would be solved in 4.5.1. Have you checked the board with 4.6 development firmware?

Turning on dshot_bidir is the best setting because it can use RPM filter. I did some tests last night, PC13 (MOTOR 4) uses TIM8_CH4N, MOTOR 4 cannot implement dshot_bidir when dshot_bitbang = OFF. This problem exists in 4.4.2-4.6.0, and firmware earlier than 4.4.2 has not been tested.

haslinghuis (approved):

haslinghuis (approved):

nerdCopter (approved):

PR #1319, opened 2024-05-14
update readme: config repo, plus fix links

nerdCopter:
* update readme announcing config repo in alert/admonition format
* fix legacy links

https://github.com/nerdCopter/unified-targets/blob/20240514_manufac_link/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,11 @@
 
 
 > [!IMPORTANT]
-
+> 
 > Unified-Targets are being sunsetted in favor of Target Configuration (config.h) files.  Please see https://betaflight.com/docs/development/manufacturer/creating-configuration for creating targets for the https://github.com/betaflight/config repository.
-
+> 
 > Instructions have been updated as Betaflight requires certification as we are migrating to a new build API.
-
+> 
 > Manufacturers can contact us on Discord to be part of our certification program.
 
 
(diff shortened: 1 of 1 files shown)


nerdCopter:
trying to use the alerts syntax, but it did not render.

> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action

haslinghuis:
Should we change 
`The instructions have been updated because Betaflight requires certification after migrating to a new build API`

nerdCopter:
> Should we change `The instructions have been updated because Betaflight requires certification after migrating to a new build API`

sure. any suggestion?

haslinghuis:
That was the suggestion - but have updated here:


`Instructions have been updated as Betaflight requires certification as we are migrating to a new build API.`
==> 
`Instructions have been updated as Betaflight requires certification after migrating to a new build API`

haslinghuis (approved):
Think this is even better :)

blckmn (approved):

PR #1324, opened 2024-05-16
Add  new target FLCO-FLYCOLORF7MICRO

tobeornottobe81:
add a new target


sugaarK:
New targets must be added to the config repository not the unified targets repository. 4.5 does not use this repository and we are not back porting new targets to the unified repository and 4.4 @tobeornottobe81

PR #1330, opened 2024-06-16
YYRC-EAGLEF405V1

tianv2mu0897:
hello:
        I would like to create a branch and merge it. I am very much looking forward to merging. If there are any issues with the branch, I hope you can provide suggestions; Thank you!


haslinghuis (commented):
- First revision should not use V1
- Betaflight no longer accepts new unified targets. Board configurations needs to be added in https://github.com/betaflight/config
- We have no YYRC manufacturer data

See https://betaflight.com/docs/development/manufacturer/requirements-for-submission-of-targets

PR #1350, opened 2024-10-20
add new target StellarF7 Dual

cvetaevvitaliy:
add new target StellarF7 Dual


haslinghuis:
This repo is here for backwards compatibility and does no longer accept new target submissions.

https://betaflight.com/docs/development/manufacturer/requirements-for-submission-of-targets

